Issue #230, opened 2017-01-13
ImportError: cannot import name 'process'


anthony-tuininga:
Originally reported by: **zaazbb (Bitbucket: [zaazbb](https://bitbucket.org/zaazbb), GitHub: [zaazbb](https://github.com/zaazbb))**

----------------------------------------

import Process from multiprocessing,  build a exe and run it, met a import error.

```
#!python

from multiprocessing import Process, Pipe, freeze_support

if __name__ == '__main__':
    freeze_support()
    ...
```

----------------------------------------
- Bitbucket: https://bitbucket.org/anthony_tuininga/cx_freeze/issue/230

anthony-tuininga:
*Original comment by* **zaazbb (Bitbucket: [zaazbb](https://bitbucket.org/zaazbb), GitHub: [zaazbb](https://github.com/zaazbb))**:

----------------------------------------

if i import Process from multiprocessing.context,  it runs ok, no import error.

anthony-tuininga:
*Original comment by* **Graham Jones (Bitbucket: [grahamrjuk](https://bitbucket.org/grahamrjuk), GitHub: [grahamrjuk](https://github.com/grahamrjuk))**:

----------------------------------------

Is this the same as #214 ?

Siecje:
I also have this problem with `cx_Freeze==5.0.1` on Windows.

`from multiprocessing.context import Process` does not work.

grahamrjuk:
I have my own (new) solution which works for me at the moment (assuming the detail is the same as in #214). I forked cx_Freeze and pushed a branch named proceess_import:
https://github.com/grahamrjuk/cx_Freeze/tree/process_import

I give no guarantee that it is the right solution.

Since it may not be the best solution I haven't raised a pull-request for it at the moment. Perhaps that is the best way to find out if I am close to a sensible solution.

Siecje:
I think creating a pull-request is a good idea, even if it doesn't get merged.

anthony-tuininga:
The problem is that the multiprocessing package has both Process (class) and process (submodule) but the imp.find_module() function on Windows finds both of them as submodules! I have pushed a change to cx_Freeze that tells it to ignore multiprocessing.Process. You can see more details on pull request #241 .

grahamrjuk:
I knew there would be a better solution than my original go. Thanks for working this out.

anthony-tuininga:
[...]

Issue #214, opened 2016-12-08
cx_freeze keeps looking in python/libs, not in compiled libs


anthony-tuininga:
Originally reported by: **Anonymous**

----------------------------------------

I `cx_freeze`d a Python program (moved from cx_freeze 4.3.4 to 5.0), and I have a problem that apparently cx_freeze keeps looking for the libs it needs in Python's (Anaconda's) directory, instead of looking for it in the directory of the executable it produced.

The current error I'm getting is the following, as I'm using `multiprocessing`:

    Traceback (most recent call last):
      File "D:\Anaconda3\lib\site-packages\cx_Freeze\initscripts\__startup__.py", line 12, in <module>
        __import__(name + "__init__")
      File "D:\Anaconda3\lib\site-packages\cx_Freeze\initscripts\Console.py", line 21, in <module>
        scriptModule = __import__(moduleName)
      File "Main.py", line 4, in <module>
      File "D:\Anaconda3\lib\multiprocessing\__init__.py", line 16, in <module>
        from . import context
      File "D:\Anaconda3\lib\multiprocessing\context.py", line 5, in <module>
        from . import process
    ImportError: cannot import name 'process'

The directory you see up there is my Python installation!!!! Why is it looking there? Now in the frozen program, if I go to: `myFrozenProgram/multiprocessing/`, I find a file called `Process.pyc`!

My setup script is the following. It's a little more complicated than usual because I add custom dlls from numpy because of a problem that happened in the past. Also the TCL part I added due to an error with tcl:

    import sys
    import os
    import json
    import glob
    from cx_Freeze import setup, Executable
    from GUI.Meta import *
    import os

    PythonPath = os.path.split(sys.executable)[0] #get python path

    os.environ['TCL_LIBRARY'] = os.path.join(PythonPath,"tcl","tcl8.6")
    os.environ['TK_LIBRARY']  = os.path.join(PythonPath,"tcl","tk8.6")

    mkl_files_json_file = glob.glob(os.path.join(PythonPath, "conda-meta","mkl-[!service]*.json"))[0] #json files that has mkl files list (exclude the "service" file)
    with open(mkl_files_json_file) as file:
        mkl_files_json_data = json.load(file)

    numpy_mkl_dlls = mkl_files_json_data["files"] #get the list of files from the json data file

    np_dlls_fullpath = list(map(lambda currPath: os.path.join(PythonPath,currPath),numpy_mkl_dlls)) #get the full path of these files

[...]

PR #241, opened 2017-02-01
Process import

grahamrjuk:
This is a solution for #214 and #230 but it probably isn't the most efficient.
It works for me at the moment and may indicate a better solution.

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -567,7 +567,7 @@ class ModuleFinder(object):
            modules are truly missing."""
         arguments = []
         importedModule = None
-        method = dis._unpack_opargs if sys.version_info[:2] >= (3, 5, 2) \
+        method = dis._unpack_opargs if sys.version_info[:3] >= (3, 5, 2) \
                 else self._UnpackOpArgs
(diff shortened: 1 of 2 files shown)


dam5h:
I also have run into this issue.  Bumping this PR to see if there are any plans to merge or evaluate alternative solutions?

anthony-tuininga:
I am closing this pull request as it doesn't actually solve the problem, unfortunately. Thankfully there is another solution which I'll explain further on.

On Windows, the imp.find_module() package is broken. It happily finds any name whatsoever as long as the file name can be loaded. If you try to do this within Python itself it will fail. For example this code "works":

```python
import imp
imp.find_module("MultiProcessing")
```
whereas this code will not
```python
import MultiProcessing
```
The correct solution is to stop using the imp module and start using the importlib package -- but that is only available in Python 3 so that will be part of cx_Freeze 6 (which will be Python 3 only).

There is a workaround for now -- and that is to tell cx_Freeze that multiprocessing.Process is not a valid module name. In the multiprocessing package, Process refers to a class and process refers to a submodule. I have pushed a change to the hooks that should solve your problem for you. I'll update the issues accordingly.

PR #245, opened 2017-02-08
Fix freezing due to import lock.

### LINKED ISSUE #242: Program freezes on threading+codecs
Calling `_codecs.lookup` in non-main thread with frozen program freezes execution:
```python
import threading, _codecs
p = threading.Thread(target=_codecs.lookup, args=('utf8',))
p.start()
p.join()
```
Program does not consume CPU and does not react to Ctrl+C.
I use python 2.7.12 (Anaconda, 64bit) and cx_freeze v5.0.1. Not reproducible with python 3.

### LINKED ISSUE #244: Program freezes when importing urllib in child thread.
The child thread will hang when it tries to import urllib while in a child thread.

```python
import threading

def child():
  print "Child importing urllib"
  import urllib
  print "Success importing urllib"

if __name__ == '__main__':
  t = threading.Thread(name=u'child', target=child)
  t.start()

  print "Joining thread"
  t.join()
  print "Done!"
```

For the above, the frozen application will hang at "Child importing urllib"

I'm using python 2.7 64 bit, Windows 7 with  cx_Freeze 5.0.1

This issue is probably similar or the same as #242

jpsnyder:
closes #244 
closes #242

### DIFF
--- a/cx_Freeze/initscripts/Console.py
+++ b/cx_Freeze/initscripts/Console.py
@@ -16,16 +16,18 @@ DIR_NAME = os.path.dirname(sys.executable)
 os.environ["TCL_LIBRARY"] = os.path.join(DIR_NAME, "tcl")
 os.environ["TK_LIBRARY"] = os.path.join(DIR_NAME, "tk")
 
-m = __import__("__main__")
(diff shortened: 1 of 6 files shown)


anthony-tuininga:
Thanks @jpsnyder

Issue #131, opened 2015-02-01
Fatal Python error: Py_Initialize: Unable to get the locale encoding ImportError: No module named 'encodings'


anthony-tuininga:
Originally reported by: **Lee Philip (Bitbucket: [lee_philip](https://bitbucket.org/lee_philip), GitHub: Unknown)**

----------------------------------------

I downloaded cx_Freeze from [here](https://bitbucket.org/anthony_tuininga/cx_freeze), installed it successfully on Ubuntu following [this thread](#32/cant-compile-cx_freeze-in-ubuntu-1304) .

After run python3 setup.py build in cx_Freeze/samples/simple,then change the dir to cx_Freeze/samples/simple/build/exe.linux-i686-3.4,run the following command ,I got the error

```
#!python

➜  exe.linux-i686-3.4  ./hello
Fatal Python error: Py_Initialize: Unable to get the locale encoding
ImportError: No module named 'encodings'
[1]    3950 abort      ./hello
➜  exe.linux-i686-3.4  
```

any idea on fixing this issue ?

----------------------------------------
- Bitbucket: https://bitbucket.org/anthony_tuininga/cx_freeze/issue/131

anthony-tuininga:
*Original comment by* **Thomas Kluyver (Bitbucket: [takluyver](https://bitbucket.org/takluyver), GitHub: [takluyver](https://github.com/takluyver))**:

----------------------------------------

The development version of cx_Freeze does not work with the system packaged Python on Debian/Ubuntu based systems (even after you work around issue #32). That is due to [this patch](http://bazaar.launchpad.net/~doko/python/pkg3.4-debian/view/head:/patches/no-zip-on-sys.path.diff), which stops Python adding `lib/pythonXY.zip` to sys.path. Executables built by cx_Freeze rely on a zip file in that location being added to sys.path.

Workaround: build executables using another Python, e.g. the Continuum built version in [Miniconda](http://conda.pydata.org/miniconda.html), or by compiling vanilla CPython yourself.

It's not clear how this could be fixed in cx_Freeze.

anthony-tuininga:
*Original comment by* **Julian Rath (Bitbucket: [jrath](https://bitbucket.org/jrath), GitHub: [jrath](https://github.com/jrath))**:

----------------------------------------

I have the some problem on win32 with py352. Building in a virtualenv

anthony-tuininga:
[...]

PR #248, opened 2017-03-05
Fixed cx_Freeze making faulty builds with Debian/Ubuntu distributed Python. (Issue #131)

peacegiverman:


anthony-tuininga:
Thanks for providing this. It gave me an idea on how to improve this permanently: namely to calculate a path and then call Py_SetPath() with the result, rather than depend on the built-in functionality which is not consistent! I hope to check this in shortly and will then close this. I'd appreciate your comments on this.

peacegiverman:
That sounds like a good idea. Making the path always consistent is a better solution.

anthony-tuininga:
Py_SetPath() is only available in Python 3 so cx_Freeze 6 will contain this functionality (it will be Python 3 only). So I'm going to defer this for bit.

anthony-tuininga:
Another solution has been implemented on both 5.x and master (6.x) branches of cx_Freeze. As mentioned earlier, thanks for the suggestions!

PR #252, opened 2017-03-10
Fixes bug caused by changes in PEP 420

mhaimes:
(https://www.python.org/dev/peps/pep-0420/)

When including a package that doesn't exist in the 'includes' list of
build options, the code that prints the error message describing the
incorrect package causes an exception.

```<module>.__path__``` is now a ```_NamespacePath``` object that does not support
```getitem()```-style indexing. It does support iterable, so this change grabs
the first element like before.

This bug is extremely simple to reproduce, simply add a package to 'includes' that doesn't exist and run the cx_Freeze setup.py script. Here is an example traceback:

```Traceback (most recent call last):
  File "setup.py", line 42, in <module>
    Executable("main.py")
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\dist.py", line 349, in setup
    distutils.core.setup(**attrs)
  File "C:\Program Files\Python36\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Program Files\Python36\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Program Files\Python36\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Program Files\Python36\lib\distutils\command\build.py", line 135, in run
    self.run_command(cmd_name)
  File "C:\Program Files\Python36\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Program Files\Python36\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\dist.py", line 219, in run
    freezer.Freeze()
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\freezer.py", line 621, in Freeze
    self.finder = self._GetModuleFinder()
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\freezer.py", line 335, in _GetModuleFinder
    package = finder.IncludeModule(name, namespace = True)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 646, in IncludeModule
    namespace = namespace)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 311, in _ImportModule
    deferredImports, namespace = namespace)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 400, in _InternalImportModule
    fp, path, info = self._FindModule(searchName, path, namespace)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 222, in _FindModule
    return None, module.__path__[0], info
TypeError: '_NamespacePath' object does not support indexing```

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -218,7 +218,7 @@ class ModuleFinder(object):
                 # Namespace package (?)
                 module = sys.modules[name]
                 info = ("", "", imp.PKG_DIRECTORY)
-                return None, module.__path__[0], info
+                return None, next(p for p in module.__path__), info
 
             # Check for modules in zip files.
             # If a path is a subdirectory within a zip file, we must have

### CODE REVIEW

**Comment by anthony-tuininga** [comment]
Can you provide a small sample that demonstrates the problem? I tried adding a module that doesn't exist to the "includes" option and got the expected error: ImportError: No module named "BadName".

**Comment by mhaimes** [comment]
Sorry! It was a long day of hacking and my description of the bug and the steps to reproduce it were completely wrong. The patch is valid, and it actually fixes a much more serious issue than just incorrect error reporting. The problem actually occurs when using the ```namespace_packages``` option. The way it falls through the error handling at line 221 of finder.py to realize that the import isn't a basic package but a namespace_package is actually broken, likely as of whenever PEP 420 went into mainstream python (I believe it was 3.3 but I could be really wrong there).

So, the steps to reproduce actually have nothing to do with an incorrect name under ```includes``` or ```packages```, but happen when you try to use ```namespace_packages```

Try the following:
1) Install zope the "old way" (into Lib/site-packages) (I got it automatically via pip install -U tornado but that really shouldn't matter)
2) make a setup.py script like:
```
build_exe_options = {
    "namespace_packages": [
        "zope",
    ],
}
```
3) when you run setup.py build, instead of cx_Freeze gracefully handling the ImportError, switching to it's namespace_packages functionality and pulling in zope, instead it spits out

```
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 217, in _FindModule
    return imp.find_module(name, path)
  File "C:\Program Files\Python36\lib\imp.py", line 296, in find_module
    raise ImportError(_ERR_MSG.format(name), name=name)
ImportError: No module named 'zope'

During handling of the above exception, another exception occurred:

  File "setup.py", line 42, in <module>
    Executable("server.py")
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\dist.py", line 349, in setup
    distutils.core.setup(**attrs)
  File "C:\Program Files\Python36\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Program Files\Python36\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Program Files\Python36\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Program Files\Python36\lib\distutils\command\build.py", line 135, in run
    self.run_command(cmd_name)
  File "C:\Program Files\Python36\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Program Files\Python36\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\dist.py", line 219, in run
    freezer.Freeze()
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\freezer.py", line 621, in Freeze
    self.finder = self._GetModuleFinder()
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\freezer.py", line 335, in _GetModuleFinder
    package = finder.IncludeModule(name, namespace = True)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 647, in IncludeModule
    namespace = namespace)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 312, in _ImportModule
    deferredImports, namespace = namespace)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 401, in _InternalImportModule
    fp, path, info = self._FindModule(searchName, path, namespace)
  File "C:\Program Files\Python36\lib\site-packages\cx_Freeze\finder.py", line 223, in _FindModule
    return None, module.__path__[0], info
TypeError: '_NamespacePath' object does not support indexing
```

when you fix the line in question:
---> line 221 change ```module.__path__[0]``` to ```next(p for p in module.__path__)```
cx_Freeze can now grab the first path properly, handle the original ```ImportError```, and pull in zope with no complaints. 

More importantly, the patch should also be backwards-compatible with when ```module.__path__``` was still a normal ```getitem```-able list. There are obviously a few ways to write this change, and if you like a different way of pulling-the-first-element-off-an-iterable, go ahead and use that instead. I just did my best to triple-check that I'm not confused, but that's always still a possibility!

File: cx_Freeze/finder.py [hunk 1]
```diff
@@ -218,7 +218,7 @@ def _FindModule(self, name, path, namespace):
                 # Namespace package (?)
                 module = sys.modules[name]
                 info = ("", "", imp.PKG_DIRECTORY)
-                return None, module.__path__[0], info
+                return None, next(p for p in module.__path__), info
```
anthony-tuininga, line 221: I am unable to replicate the issue currently. The following code should work as well. Can you check?

    return None, next(module.__path__), info

Thanks.

mhaimes: sigh, I wish that fixed it...

```_NamespacePath``` objects are iterators but NOT generators. In general, my understanding of the python standard-library design philosophy is that they use generators for ephemeral/in-flight data, and iterators for static, iterable data that you might need to save and iterate over more than once. next() only works on fully-qualified generators, and the p for p trick just wraps an iterable object with a brand-new generator.

My solution is certainly a bit clunky, but I couldn't figure out how to get the "first" element off an iterator without flushing the whole iteration. i.e., you could do something like ```list(module.__path__)[0]``` but then you potentially copy the entire list of paths when you only want one, and I figured anything to keep cx_Freeze builds as fast as possible is worth it.

For the record, the code sample you suggested generates the following error instead (I just tested):
```
  return None, next(module.__path__), info
TypeError: '_NamespacePath' object is not an iterator
```

As far as helping you to reproduce it, my setup is pretty vanilla.. windows 10 64-bit, python3.6 installed off the vanilla python installer for windows/amd64, and cx_Freeze installed via pip

adrientetar: Does next(iter(module.__path__)) not work here?

55octet: I am wondering if anyone was able to reproduce this issue.  I am currently experiencing the same namespace package issue.  None of these fixes seem to correct the issue.  I am trying to import custom namespaces that I have written, and it definitely isn't finding them either.  I am running on RHEL 7.3, Python 3.4.2, and cx_Freeze 5.0.1.

My __init__.py files are using the standard pre-PEP420 import for setuptools:

__import__('pkg_resources').declare_namespace(__name__)

The __init__.py files have been removed from the site-packages directories, and they play nicely with pip.

anthony-tuininga: Can you provide the script that causes this problem to occur? I'd like to take a closer look. Thanks.

anthony-tuininga:
An alternative solution to this has now been merged. If that doesn't solve your issue, please let me know!

Issue #254, opened 2017-03-29
cx_freeze 5.0.1 and python 3.6.1 - bug at freezing


FreeHarry:
I got a bug at the freezing my app. It appears if the code is trying to get a value from the list `arguments`, but the list is still empty.

Replace the line 561 in finder.py ...
`fromList, = arguments`

with ...
`fromList = arguments[0] if arguments else []`

does fix the bug for me.

```
running build_exe
Traceback (most recent call last):
  File "setup_cx.py", line 85, in <module>
    executables = [executable_1],
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\dist.py", line 349, in setup
    distutils.core.setup(**attrs)
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\dist.py", line 219, in run
    freezer.Freeze()
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\freezer.py", line 623, in Freeze
    self._FreezeExecutable(executable)
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\freezer.py", line 199, in _FreezeExecutable
    finder.IncludeFile(exe.script, exe.moduleName)
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\finder.py", line 633, in IncludeFile
    deferredImports)
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\finder.py", line 474, in _LoadModule
    self._ScanCode(module.code, module, deferredImports)
  File "C:\Users\hh16534\AppData\Local\Programs\Python\Python36-32\lib\site-packages\cx_Freeze\finder.py", line 564, in _ScanCode
[...]

PR #260, opened 2017-04-26
Fix variable assigning on an empty list

FreeHarry:
See issue #254 for details...


anthony-tuininga:
Do you have an example that demonstrates the problem for issue #254? I'd like to make sure that this solution doesn't just address symptoms and ignores the root cause. Thanks.

Issue #212, opened 2016-12-05
win32file needs win32timezone in python 3


anthony-tuininga:
Originally reported by: **Harald Hanöfner (Bitbucket: [HarryH](https://bitbucket.org/HarryH), GitHub: [HarryH](https://github.com/HarryH))**

----------------------------------------

The module ```win32file``` from the pywin32 package needs the module ```win32timezone``` for many time relevant methods when using python 3. This module isn't automatically added at freezing.

My recommendation is to add the following line to the hook ```load_win32file``` within the ```hooks.py```:

```
#!python
finder.IncludeModule("win32timezone")

```
It works well for me.

----------------------------------------
- Bitbucket: https://bitbucket.org/anthony_tuininga/cx_freeze/issue/212

anthony-tuininga:
Merge request resolving this issue has been merged.

Issue #207, opened 2016-10-10
sqlite3.dll not shipped


anthony-tuininga:
Originally reported by: **Johannes Eckstein (Bitbucket: [skidzo](https://bitbucket.org/skidzo), GitHub: [skidzo](https://github.com/skidzo))**

----------------------------------------

Hi all the last bug i solved after installing cx_Freeze on Python35 is worth to be mentioned I guess.

To get cx_Freeze working again:
https://github.com/sekrause/cx_Freeze-Wheels

The sqlite3.dll somehow gets lost...

Python35\DLLs\sqlite3.dll, this may be because the depending file is _sqlite3.pyd in the same Directory. Somehow this dependency does not get resolved.

This Problem affects various other packages as tkinter and so on.

Nevertheless I got my PyQT5.7 application running!

Good luck in the new Job and have fun!

Thanks a lot for this package!

Keep up the good work!

----------------------------------------
- Bitbucket: https://bitbucket.org/anthony_tuininga/cx_freeze/issue/207

anthony-tuininga:
*Original comment by* **Parry Cadwallader (Bitbucket: [wbc35](https://bitbucket.org/wbc35), GitHub: Unknown)**:

----------------------------------------

@skidzo did you have to do anything else after installing the wheel?  I upgraded to cx_freeze 5.0 and I'm still receiving the `_sqlite3` DLL not found issue.

anthony-tuininga:
*Original comment by* **David Hobbs (Bitbucket: [hobbsd](https://bitbucket.org/hobbsd), GitHub: [hobbsd](https://github.com/hobbsd))**:

----------------------------------------

I had the same problem and got the sqlite3.dll included by adding it to the `include_files` under `build_exe` in the setup.py `options`. For example:

    options = {'build_exe': {'include_files': [os.path.join(sys.base_prefix, 'DLLs', 'sqlite3.dll']}}

anthony-tuininga:
*Original comment by* **Harald Hanöfner (Bitbucket: [HarryH](https://bitbucket.org/HarryH), GitHub: [HarryH](https://github.com/HarryH))**:

----------------------------------------

Since the module sqlite3 belongs to the python standard library I would recommend to add a new hook within the cx_Freeze library.

Adding the following code snippet to the ```hooks.py``` works well for me:
```
#!python
def load_sqlite3(finder, module):
    """In Windows, the sqlite3 module requires an additional dll sqlite3.dll to
[...]

PR #261, opened 2017-04-26
New hook for sqlite3 on Win and improvement for win32file hook

FreeHarry:
See issue #207 and #212 ...

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -677,9 +677,10 @@ def load_win32com(finder, module):
 
 
 def load_win32file(finder, module):
-    """the win32api module implicitly loads the pywintypes module; make sure
-       this happens."""
+    """the win32api module implicitly loads the pywintypes and win32timezone
+       module; make sure this happens."""
     finder.IncludeModule("pywintypes")
+    finder.IncludeModule("win32timezone")
 
 
 def load_wx_lib_pubsub_core(finder, module):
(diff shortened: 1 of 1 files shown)


cx_Freeze/hooks.py
```diff
@@ -677,9 +677,10 @@ def load_win32com(finder, module):
 
 
 def load_win32file(finder, module):
-    """the win32api module implicitly loads the pywintypes module; make sure
-       this happens."""
+    """the win32api module implicitly loads the pywintypes and win32timezone
+       module; make sure this happens."""
     finder.IncludeModule("pywintypes")
+    finder.IncludeModule("win32timezone")
```
anthony-tuininga, line 683: This change is not needed as far as I can tell. I just tried with a new installation of Python 3.6 and it works just fine without win32timezone. Do you have an example that demonstrates the need?

FreeHarry: Here is an example for you :-)
I modified minor things of the wxPython example supplied with the cx_freeze package. You can compare these for seeing the difference.

If you run the setup.py for this example you will first see the error described in issue #254. If you patch cx_freeze with the pull request #260, the script runs perfect.

After freezing, start the executable and press the button "Push me". Now you will get the error described in the issue #212 => `ModuleNotFoundError: No module named 'win32timezone'`

My environment: Win 7/10 64Bit, Python 3.6.1, pywin32 and wxPython 4.0.0a1.
Attached example: [cx_freeze-bug_test.zip](https://github.com/anthony-tuininga/cx_Freeze/files/1005069/cx_freeze-bug_test.zip)

anthony-tuininga: Got it. Thanks!

cx_Freeze/hooks.py
```diff
@@ -814,3 +815,12 @@ def load_clr(finder, module):
...
     dllname = 'Python.Runtime.dll'
     finder.IncludeFiles(os.path.join(module_dir, dllname), dllname)
+    
+    
+def load_sqlite3(finder, module):
+    """In Windows, the sqlite3 module requires an additional dll sqlite3.dll to
+       be present in the build directory."""
+    if sys.platform == "win32":
+        dll_name = "sqlite3.dll"
+        dll_path = os.path.join(sys.base_prefix, "DLLs", dll_name)
```
dandersson, line 825: `sys.base_prefix` breaks on Python 2.7, since `sys.base_prefix` is only present in Python 3.3+ ([see docs](https://docs.python.org/3/library/sys.html#sys.base_prefix)). In older Python installations, one should look for `sys.real_prefix` which is set by `virtualenv` within virtual environments, and fall back to `sys.prefix` which is always set (but does not point to the system installation if run from a virtual environment).

See pull request #272.

PR #267, opened 2017-05-09
Add FAQ entry for command prompt flashing up

takluyver:
This is a common failure mode, so it's worth explaining what's happening and how to get past it.


anthony-tuininga:
Thanks, Thomas!

PR #279, opened 2017-06-28
Do not create version resource when version is omitted

### LINKED ISSUE #154: Error in freezer.py
Originally reported by: **Anonymous**

----------------------------------------

I have not seen this error posted anywhere else, though I would like to bring it to your attention. I was using cx_freeze greatly, when an unforeseen error came up in the cmd prompt (when converting to .exe). It traced the problem to **line 788 of *freezer.py***. This lines code is: **parts = version.split()** When I did some digging into this, I found that the **version.split()** command was not working and thus crashing build_exe. I have since edited this line of code to be **parts = '1234'** and it seems to be working fine! Wonderful program and I hope this helps to develop it. 

System Information: OS - Windows 10 | Python Version: 3.4.1

----------------------------------------
- Bitbucket: https://bitbucket.org/anthony_tuininga/cx_freeze/issue/154

rcutmore:
Previously an `AttributeError` would be thrown during the freeze process when pywin32 was installed and the version was not specified. This change avoids throwing an error by just specifying a default version of "0.0.0.0" if needed. If we would rather not specify a default version it would be nice to throw a `ConfigError` that indicates the version is missing instead.

This fixes #154. I searched and couldn't find any other posted issues related to this.

Edit: It might be better to just return early in `_AddVersionResource` if the version is missing. Either way I'm happy to revise this PR for any changes.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -149,11 +149,16 @@ class Freezer(object):
         self._VerifyConfiguration()
 
     def _AddVersionResource(self, exe):
+        warning_msg = "*** WARNING *** unable to create version resource"
         try:
             from win32verstamp import stamp
         except:
-            print("*** WARNING *** unable to create version resource")
+            print(warning_msg)
             print("install pywin32 extensions first")
+            return
+        if not self.metadata.version:
+            print(warning_msg)
+            print("version must be specified")
             return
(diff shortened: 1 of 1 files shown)


rcutmore:
I updated this to avoid creating a version resource when the version number is omitted.

PR #286, opened 2017-07-13
[WIP] Fix No module named __startup__

klensy:
Should fix issue, please try ```pip install git+https://github.com/klensy/cx_Freeze.git@fix_209```.
Lightly tested on windows, linux; with\without venv.

Don't merge, need test.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -494,13 +494,6 @@ class Freezer(object):
         finder.ReportMissingModules()
 
         targetDir = os.path.dirname(fileName)
-        if sys.platform != "win32":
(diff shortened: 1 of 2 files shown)


anthony-tuininga:
I have a different solution for Python 6.x (Python 3.x only) so this might be a reasonable solution for cx_Freeze 5.x. How goes the testing on this? Is this ready to merge in your opinion? It looks reasonable to me.

bobatsar:
Just wanted to note that this PR fixes the issue on my side too on windows with venv.

anthony-tuininga:
Since it looks reasonable and @bobatsar has indicated it works on Windows as well, I'm going to go ahead and merge this.

PR #304, opened 2017-08-31
Skip unknown shared library files which are returned from otool

bobatsar:
When freezing numpy with current v5.x branch, libnpymath.a references e.g. ../Contents/MacOS/lib/numpy/core/lib/libnpymath.a(npy_math.o):
which is itself or an object inside libnpymath.a

The issue is that those paths are not correctly split (no space in between) but it also references itself so references to unknown files are just skipped.

### DIFF
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -133,7 +133,6 @@ class bdist_mac(Command):
             files.extend([os.path.join(root, f).replace(self.binDir + "/", "")
                           for f in dir_files])
         for fileName in files:
-
             # install_name_tool can't handle zip files or directories
             filePath = os.path.join(self.binDir, fileName)
             if fileName.endswith('.zip'):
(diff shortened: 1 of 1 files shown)


cx_Freeze/macdist.py
```diff
@@ -133,7 +133,6 @@ def setRelativeReferencePaths(self):
             files.extend([os.path.join(root, f).replace(self.binDir + "/", "")
                           for f in dir_files])
         for fileName in files:
-
```
anthony-tuininga, line 136: Please don't make arbitrary white space changes in pull requests. Can you revert this and the other one?

cx_Freeze/macdist.py
```diff
@@ -154,10 +153,19 @@ def setRelativeReferencePaths(self):
...
+                # files, like
+                # ".../Contents/MacOS/lib/numpy/core/lib/libnpymath.a(npy_math.o):"
+                # and the file can not be located (in this case the
+                # libnpymath.a references itself)
+                if not os.path.exists(referencedFile):
+                    print("ERROR: file {} does not exist. Skipping it".format(
+                        referencedFile))
+                    continue
+
                 if referencedFile.startswith('@executable_path'):
```
anthony-tuininga, line 169: If the referencedFile starts with @executable_path, is the code you just added going to cause it to get skipped? Perhaps your code block is in the wrong place?

cedk: Indeed if the referencedFile starts with '@executable_path', it was skipped any way. But it will be better to not have the 'ERROR' print for this case, so the block quote should be moved after the '@executable_path' test.

cedk:
I have also this issue when bundling a GTK application with lib/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-ani.a

PR #307, opened 2017-09-06
Fixing issue with @rpath causing file copy errors on bdist_mac and pyqt5

voglster:
I ran into an issue with pyqt5 installed in an anaconda virtual env using python 3.6 
Somehow dylibs get pulled in with @rpath in their path.  After hardcoding the path in macdist I realized that I could just pass in a @rpath replacement from setup.py.

Couple caveats... this has been tested on my machine only in anaconda and python 3.6. I am very new to mac (normally just use linux) but the change seems rather harmless as copy_file will just fail when it gets a @rpath prefixed filename but I definitely do not know all the ramifications downstream.

### DIFF
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -1,4 +1,4 @@
-from distutils.core import Command
+from distutils.core import Command, DistutilsFileError
 import os
 import plistlib
 import stat
@@ -89,6 +89,7 @@ class bdist_mac(Command):
             '--deep option.'),
         ('codesign-resource-rules', None, 'Plist file to be passed to ' \
             'codesign\'s --resource-rules option.'),
+        ('rpath-lib-folder', None, 'replace @rpath with given folder for any files')
     ]
 
     def initialize_options(self):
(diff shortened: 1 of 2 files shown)


cx_Freeze/macdist.py
```diff
@@ -171,9 +176,12 @@ def setRelativeReferencePaths(self):
                 if (name not in files and not path.startswith('/usr') and not
                         path.startswith('/System')):
                     print(referencedFile)
-                    self.copy_file(referencedFile,
-                                   os.path.join(self.binDir, name))
-                    files.append(name)
+                    try:
+                        self.copy_file(referencedFile, os.path.join(self.binDir, name))
+                    except DistutilsFileError as e:
+                        print(f"issue copying {referencedFile} to {os.path.join(self.binDir, name)} error {e} skipping")
```
anthony-tuininga: f-strings are new in Python 3.6. Can you replace this with code that works in Python 2.7, 3.5 and 3.6?

voglster: sure thing... didn't think about backward compatibility for F strings

voglster:
I believe this will work but unfortunately I did not have time to test on older python versions.

anthony-tuininga:
It looks okay so I have merged it. Thanks!

PR #314, opened 2017-09-22
PEP8 complient method naming

tjallingt:
Altered capitalization of methods called on win32 service handler to match PEP8 recommendations


anthony-tuininga:
It might have been better to use PEP 8 compliant naming at the beginning....but now that it has been there for a while, changing it is going to cause problems for those already using it!

marcelotduarte:
I am accepting your code and I will make adaptations to maintain compatibility. Thanks!

PR #347, opened 2018-01-09
print() is a function in modern Python

cclauss:


anthony-tuininga:
Thanks, but that file is actually obsolete. Thanks for bringing it to my attention, however!

PR #365, opened 2018-03-02
Always copy dependent files to root directory on darwin

cedk:
cebdef5f65e260b3425755e356208b818806af97 changed this behavior but it is requires to build on macOS.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -191,6 +191,10 @@ class Freezer(object):
         self.filesCopied[normalizedTarget] = None
         if copyDependentFiles \
                 and source not in self.finder.excludeDependentFiles:
+            # Always copy dependent files on root directory
+            # to allow to set relative reference
+            if sys.platform == 'darwin':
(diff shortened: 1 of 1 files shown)


cx_Freeze/freezer.py
```diff
@@ -191,6 +191,10 @@ def _CopyFile(self, source, target, copyDependentFiles,
         self.filesCopied[normalizedTarget] = None
         if copyDependentFiles \
                 and source not in self.finder.excludeDependentFiles:
+            # Always copy dependent files on root directory
+            # to allow to set relative reference
+            if sys.platform == 'darwin':
+                targetDir = self.targetDir
```
anthony-tuininga, line 197: Can you provide some additional information on why this is needed?

cedk: Here: https://github.com/anthony-tuininga/cx_Freeze/blob/master/cx_Freeze/macdist.py#L195
The reference path is always set from @executable_path with only the file name.

Issue #366, opened 2018-03-12
cx_Freeze 5.1.1: Lots of duplicate .dll files.


ltjax:
Hello. I'm freezing a simple application with PyQt5, numpy, scipy and matplotlib.  Lots of dlls from the libs are duplicated many times. For example, api-ms-win-crt-runtime-l1-1-0.dll is in there 19 times, python36.dll 16 times, QtCore4.dll 4 times. Some of these files are huge, making the folder ~600mb for an about 350locs app.
It seems to just add each .dll for each entry in the lib/ folder instead of consolidating them.

reubengann:
I second this. My project has matplotlib, numpy, pandas, and pyqt. I don't expect the application to be small, but if possible I'd like it not to include python36.dll 21 times.

cainesi:
I am having the same issues on Windows.  It seems like the extra copies are being produced by the ._CopyFile method on the Freezer object.  When that method is called to copy a file, it in turn calls the ._GetDependentFiles method to find other dependent files to also copy.  The dependents are being determined (I think?) by calling some native windows function to determine the bindings.  If it locates dependent libraries then they are also copied into the same directory.  Therefore if there are a bunch of libraries in the application in different locations that all depend (e.g.) on Qt5Core.dll, then it looks like Qt5Core.dll gets copied to each of the locations.  

Could we prevent this by tracking which libraries have already been included in the application as dependents elsewhere?  I don't know enough about how linking works on Windows to know if that would cause problems...

tigarmo:
I'm also having this issue. As far as I can tell this change was added in issue #328 , but that fix seems overkill for the majority of the applications; my test script that just uses subprocess has `python36.dll` and `VMRUNTIME140.dll` in both the root freeze dir and the `lib` dir.

Is there a workaround for this new behavior of copying dependent files to the same directory as the source file?

cainesi:
Do we know why some libraries needed to be co-located in the same directory?

tigarmo:
I don't think it's the case that the libraries need to be next to each other necessarily, but the dependencies have to be "findable" at runtime. So when the change to put some files in a `lib` directory was made it probably broke this lookup in some cases.

For our use case the size of the frozen directory went from 900 MB to 1.8 GB from all the duplicate libs, which complicates releases. Is there anything we can do to help out?

cainesi:
[...]

PR #400, opened 2018-07-16
Added an option to put all dependences in one place.  Addresses issues #366.

cainesi:
Added a new parameter "small_app" to the Freezer object that, if true: 1) causes all dependencies to be copied into the top directory of the application (rather than the directory of the file that they are dependency of), and 2) causes ConsoleSmallApp.py to be used as the init script in place of Console.py.  ConsoleSmallApp.py is the same as Console.py, except that it causes the application's directory to be added to the PATH environment variable.  Addresses issue in #366.


rockferd:
Are there any plans to release this change?

cainesi:
> Are there any plans to release this change?

I am guessing not--the change is a bit ugly I think.  But you can just download the branch if you want to use it...

marcelotduarte:
@anthony-tuininga and @cainesi 
The question about lot of files, I've been working on this, and have a few more things to do, especially for Windows, but I think we might have something minimal like what I did on PR #518.
We can close this PR?

anthony-tuininga:
Yes, I think we can close this for now. @cainesi can update it or create a new one if he disagrees!

cainesi:
> Yes, I think we can close this for now. @cainesi can update it or create a new one if he disagrees!

Whatever you think is best.  I will be interested in seeing a real solution when available, so I can stop using my kludgy patch.  Thanks for your work on this!

PR #412, opened 2018-08-07
Improvements

diorcety:
Remove prefix hack on Windows: Maybe not the good solution, but it's not working in virtualenv. Have to check if we are in virtualenv?

Don't fail on _AddVersionResource : If the project version is not digit.digit.digit.digit, it's blocking. Forcing to uninstall pywin32 or to use another version value

Support missing IMPORT_FROM: the from . import XXXX is not find by the Finder. This is easy to check, try to cxfreeze a project using pyreadline the "from . import release" inside "pyreadline/rlmain.py is not handled, then at execution this module is missing

Add Administrator executable versions : New base values: ConsoleAsAdministrator and Win32GUIAsAdministrator. This allows on Windows to have an UAC escalation when running the program

### DIFF
--- a/cx_Freeze/dist.py
+++ b/cx_Freeze/dist.py
@@ -252,16 +252,6 @@ class install(distutils.command.install.install):
         self.install_exe = None
 
     def finalize_options(self):
-        if self.prefix is None and sys.platform == "win32":
-            try:
-                import winreg
-            except:
-                import _winreg as winreg
-            key = winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE,
-                    r"Software\Microsoft\Windows\CurrentVersion")
-            prefix = str(winreg.QueryValueEx(key, "ProgramFilesDir")[0])
-            metadata = self.distribution.metadata
-            self.prefix = "%s/%s" % (prefix, metadata.name)
         distutils.command.install.install.finalize_options(self)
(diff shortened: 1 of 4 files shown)


anthony-tuininga:
> Remove prefix hack on Windows: Maybe not the good solution, but it's not working in virtualenv. Have to check if we are in virtualenv?

A better explanation of why this is an improvement would be useful.

> Don't fail on _AddVersionResource : If the project version is not digit.digit.digit.digit, it's blocking. Forcing to uninstall pywin32 or to use another version value

I'm not sure ignoring errors is an improvement! Why do you need to use a version that isn't supported?

> Support missing IMPORT_FROM: the from . import XXXX is not find by the Finder. This is easy to check, try to cxfreeze a project using pyreadline the "from . import release" inside "pyreadline/rlmain.py is not handled, then at execution this module is missing

Can you provide a small sample that demonstrates the issue? That way it can be included in the tests -- which definitely need to be beefed up!

> Add Administrator executable versions : New base values: ConsoleAsAdministrator and Win32GUIAsAdministrator. This allows on Windows to have an UAC escalation when running the program

Instead of including all of these in one pull request, can you create four of them? One for each one? That makes it easier to accept the ones that aren't controversial and allows people to comment on them as well.

Issue #368, opened 2018-03-15
Windows executable icon file update issue


omidgl:
in freezer.py>_FreezeExecutable, file permission check and update is done after updating the icon file, so you get " "The media is write protected" error if your python/cx_freeze is located on a read-only location like network drive.

PR #430, opened 2018-10-09
Make executable writable before adding icon. Fixes issue #368

arossmanith:
Setting the application icon fails if the base executable is copied from a write-protected origin, see issue #368. I moved the code setting the write permission bit up a few lines to prevent this from happening.


anthony-tuininga:
Thanks!

PR #435, opened 2018-11-08
print error when cx_Freeze.util.GetDependentFiles(path) produce cx_Freeze.util.BindError

buxx:
Hello,

I had some issues on build process on client computer (Windows). It's user name was including a `é` character and `dependentFiles = cx_Freeze.util.GetDependentFiles(path)` was raising an `cx_Freeze.util.BindError` error.

I had to debug on computer client to find error because no log is produced in that case. So i suggest to add a log like in this PR.

Is that okay for you ?


anthony-tuininga:
Thanks!

PR #440, opened 2018-11-27
Add base for Win32 GUI subsystem, with GPU enablement exports (v5.x branch)

fr00b0:
This pull request adds a new base executable that is the same as the `Win32GUI` base executable, but with the addition of two exported symbols:

```cpp
__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
__declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
```

These symbols will cause video drivers from Nvidia and AMD respectively to automatically select discrete high performance graphics if the machine has a integrated/discrete hardware combo.

I noticed there was an issue (anthony-tuininga/cx_Freeze#363) about this in your repo, so I'd thought I'd ask if there was any interest in upstreaming my solution to this problem. This is useful for a number of application categories, including games.

__Please note:__ This pull request is based on the v5.x branch (python 2.7) since that is what I needed. If this is something that the `cx_Freeze` would consider merging, I would be willing to create a pull request towards the master branch as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,9 @@ class build_ext(distutils.command.build_ext.build_ext):
             compiler_type = self.compiler.compiler_type
             if compiler_type == "msvc":
                 extraArgs.append("/MANIFEST")
+                if ext.name == "cx_Freeze.bases.Win32GUI_GPU":
+                    implib = os.path.join(self.build_temp, ext.name)
+                    extraArgs.append("/IMPLIB:{}".format(implib))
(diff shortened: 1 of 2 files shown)


illume:
Thanks :)

René from pygame here. Yes, I am interested in this as an option too.

illume:
This would need documentation.

anthony-tuininga:
Can you recreate this for merge into master instead? I have no plans to continue supporting v5.x at this point.

PR #444, opened 2018-12-04
fix typo that causes NameError when freezing pythonet

danielhrisca:


anthony-tuininga:
There were merge conflicts so I made the minor changes myself. Thanks!

PR #445, opened 2018-12-05
pymunk could not find the .dll/.so/.dylib file.

illume:
This is a hook for pymunk.


illume:
This makes the .dll be copied in twice. I need to find another solution (which I think is changing the pymunk code to look elsewhere for the dll).

PR #446, opened 2018-12-09
lextab yacctab files not included in pycparser with cx_freeze.

illume:
These two files from pycparser are not included with programs built with cx_freeze.
The problem is on windows, it then tries to create the files in the current directory... which is often not permitted.

Related: 
- https://bitbucket.org/cffi/cffi/issues/392/lextab-yacctab-not-included-with-cx_freeze
- https://github.com/viblo/pymunk/issues/151


anthony-tuininga:
Thanks!

illume:
Thank you for cx_Freeze :)

PR #468, opened 2019-04-28
Add -municode flag when building on Windows

zegervdv:
patch applied from https://github.com/msys2/MINGW-packages/blob/4c18633ba2331d980f00aff075f17135399c43c5/mingw-w64-python-cx_Freeze/0001-fix-build.patch

Credit for the patch goes to the msys2/MINGW-packages maintainers.
I only applied it so it could install the latest HEAD version of cx_freeze via pip.


anthony-tuininga:
Thanks!

PR #476, opened 2019-07-11
include module unicodedata

marcelotduarte:
_socket extension on python 3.7 (and possible previous versions) uses encondings.idna and that requires unicodedata
This resolves stranger crashes on sockets (windows and linux)
Initially I implemented a load__socket hook like the py2exe project, but I think that globally we can avoid bug in other modules.


anthony-tuininga:
Thanks!

PR #477, opened 2019-07-11
the asyncio must be loaded as a package.

marcelotduarte:
asyncio as module inserts only __init__ on library; with this hooks resolves all issues


anthony-tuininga:
Thanks!

PR #479, opened 2019-07-11
_cffi_backend module requires libffi on a specific directory

marcelotduarte:
I am getting an ImportError when working with the cryptography packages.
Only on linux.
This patch has been rewritten and depends on #494

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -86,6 +86,19 @@ def load_ceODBC(finder, module):
     finder.IncludeModule("decimal")
 
 
+def load__cffi_backend(finder, module):
+    """_cffi_backend module requires libffi on a specific directory"""
+    if os.name == 'posix':
+        module_dir = os.path.dirname(module.file)
+        skip_count = len(module_dir)
+        for dirpath, _, filenames in os.walk(os.path.join(module_dir, ".libs_cffi_backend")):
+            for name in filenames:
+                source_path = os.path.join(dirpath, name)
+                target_path = os.path.join("lib", source_path[skip_count + 1:])
+                finder.IncludeFiles(source_path, target_path)
+        finder.ExcludeDependentFiles(module.file)
+
+
 def load_cx_Oracle(finder, module):
     """the cx_Oracle module implicitly imports datetime; make sure this
        happens."""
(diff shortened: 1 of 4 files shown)


cx_Freeze/hooks.py
```diff
@@ -71,6 +71,11 @@ def initialize(finder):
         finder.ExcludeModule("vms_lib")
     finder.ExcludeModule("new")
     finder.ExcludeModule("Tkinter")
+    #include the _cffi_backend (only if present in system or venv)
+    try:
+        finder.IncludeModule("_cffi_backend")
+    except ImportError:
+        pass
```
anthony-tuininga, line 78: I'm not excited about always including _cffi_backend. This should be added to the relevant module/package hooks instead, I think -- not global!

anthony-tuininga: I just created a hook for cryptography that does this and it seems to work. Please take a look. Thanks!

marcelotduarte: I would not have liked to make this patch like this. But I'll try to explain to you why I chose to do it that way. If you check it out, I made some modifications until I get to this "simpler" version.
But first I will talk about cffi: cffi is a package for compiling modules into other packages. So we'll take cryptography as an example, and it needs cffi to generate some modules, but the modules generated don't need the cffi package, just the _cffi_backend module, which in windows is _cffi_backend.pyd.
So the patch you made solves, but includes an unnecessary package, cffi, and could include only the _cff_backend modules.
But now I will talk because I chose to do the way I did: I use some packages that use cffi: cryptography, bcrypt, Pillow ... so for each case, there would have to be a hook. So if you have 1000 packages using cffi, there will always be an issue about it. Adding globally, I solved the problem not only for all 3 packages, but for any package that uses cffi. And there is one more thing. If the package is not in the system, ie if you do not have a package that uses cffi in venv, it will not be forced to be included.

marcelotduarte: Oh, and in linux, the patch #494 is necessary to include libffi*.so and others in proper path.

anthony-tuininga: I've merged #494 and modified the hook for cryptography as you suggested. Thanks! I'm not sure how this PR got closed, but I think its resolved for now. :-)

marcelotduarte: ok, then I'll hook to other known libraries then,

PR #482, opened 2019-07-11
pytz module requires zoneinfo data included in library.zip.

marcelotduarte:
I did several tests, it does not matter to include in the directory, have to put in the zip file.

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -135,6 +135,7 @@ class ModuleFinder(object):
         self._builtinModules = dict.fromkeys(sys.builtin_module_names)
         self._badModules = {}
         self._zip_modules_cache = ZipModulesCache()
+        self.zipIncludes = []
         cx_Freeze.hooks.initialize(self)
         initialExcludedModules = self.excludes.copy()
         self._AddBaseModules()
@@ -644,6 +645,10 @@ class ModuleFinder(object):
             sys.stdout.write("This is not necessarily a problem - the modules "
                              "may not be needed on this platform.\n")
             sys.stdout.write("\n")
+
+    def ZipIncludeFiles(self, sourcePath, targetPath):
+        """Include the files in the given library.zip in the target build."""
+        self.zipIncludes.append((sourcePath, targetPath))
 
 
 class Module(object):
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -619,6 +619,8 @@ class Freezer(object):
         # write any files to the zip file that were requested specially
         for sourceFileName, targetFileName in self.zipIncludes:
             outFile.write(sourceFileName, targetFileName)
+        for sourceFileName, targetFileName in self.finder.zipIncludes:
+            outFile.write(sourceFileName, targetFileName)
 
         outFile.close()
 
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -559,6 +559,18 @@ def load_PyQt5_QtPrintSupport(finder, module):
     name, QtCore = _qt_implementation(module)
     copy_qt_plugins("printsupport", finder, QtCore)
 
+
+def load_pytz(finder, module):
+    """pytz module requires zoneinfo data included in library.zip."""
+    module_dir = os.path.dirname(module.file)
+    skip_count = len(os.path.dirname(module_dir))
+    for dirpath, _, filenames in os.walk(os.path.join(module_dir, "zoneinfo")):
+        for name in filenames:
+            source_path = os.path.join(dirpath, name)
+            target_path = source_path[skip_count + 1:].replace('\\', '/')
+            finder.ZipIncludeFiles(source_path, target_path)
+
+
 def load_reportlab(finder, module):
     """the reportlab module loads a submodule rl_settings via exec so force
        its inclusion here"""


anthony-tuininga:
Do you have a sample script that demonstrates the problem that you can share? Perhaps include it in the samples directory? There already is the ability to include files in the zip file and it is called zipIncludes already, too. So I'd like to make sure that this patch is necessary first. Thanks!

marcelotduarte:
Previously I did like this, included in the project using zipIncludes ... but as the pytz library is very common, I decided to add it automatically.
I will make an example later.

anthony-tuininga:
@marcelotduarte, I tried the sample you provided without the patch and it worked just fine! Can you create a sample that doesn't work without the patch so I can see why this patch is required? Thanks!

marcelotduarte:
If you receive a warning, it is failing. You'll see the date in utc, not in your locale.

anthony-tuininga:
I did not receive a warning. It showed the time in my locale without any problems! This is with Python 3.7 on Linux.

marcelotduarte:
See my output on windows:
(pytz) M:\github\cx_Freeze\cx_Freeze\samples\pytz>pip list
Package    Version
---------- -------
cx-Freeze  6.0
pip        19.3.1
pytz       2019.3
setuptools 41.2.0
tzlocal    2.0.0

(pytz) M:\github\cx_Freeze\cx_Freeze\samples\pytz>python test_pytz.py
Sat, 28 Dec 2019 01:31:30 -0300

(pytz) M:\github\cx_Freeze\cx_Freeze\samples\pytz>build\exe.win32-3.7\test_pytz.
exe
WARNING: fail to load pytz timezones
Sat, 28 Dec 2019 04:31:51 +0000

(pytz) M:\github\cx_Freeze\cx_Freeze\samples\pytz>

anthony-tuininga:
Are you able to test this on Linux as well? That way we can know if it is Windows specific. I don't use Windows except for testing, so would have to fire up the virtual machine to check this out -- so that would have to wait until some time later.

marcelotduarte:
On linux it is ok.

marcelotduarte:
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ python test_pytz.py 
Sat, 28 Dec 2019 01:34:13 -0300
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ build/exe.
exe.linux-x86_64-3.7/ exe.win32-3.7/        
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ build/exe.linux-x86_64-3.7/test_pytz 
Sat, 28 Dec 2019 01:34:29 -0300

marcelotduarte:
Interestingly, I had previously tested on linux ubuntu 19.04 when I made the patch. Now I'm with 19.10 ...

anthony-tuininga:
Ok. I'll have to find some time to fire up the virtual machine -- unless you can devise a test that fails on Linux, of course!

marcelotduarte:
Excuse me, in my linux it pass with cx-freeze from my base.
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ pip list 
Package       Version
------------- -------
cx-Freeze     6.0    
pip           19.3.1 
pkg-resources 0.0.0  
pytz          2019.3 
setuptools    42.0.2 
tzlocal       2.0.0  
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ python test_pytz.py 
Sat, 28 Dec 2019 01:52:52 -0300
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ build/exe.linux-x86_64-3.7/test_pytz 
WARNING: fail to load pytz timezones
Sat, 28 Dec 2019 04:52:56 +0000

marcelotduarte:
And fails with:
pip install -U git+https://github.com/anthony-tuininga/cx_Freeze.git

(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ pip list 
Package       Version
------------- -------
cx-Freeze     6.1    
pip           19.3.1 
pkg-resources 0.0.0  
pytz          2019.3 
setuptools    42.0.2 
tzlocal       2.0.0  
(pytz) marcelo@teste6:~/M/github/cx_Freeze/cx_Freeze/samples/pytz$ build/exe.linux-x86_64-3.7/test_pytz 
WARNING: fail to load pytz timezones
Sat, 28 Dec 2019 04:55:36 +0000

anthony-tuininga:
So pytz does allow for the environment variable ``PYTZ_TZDATADIR`` to be set pointing to the directory in which the timzone information is found. This was done for matplotlib so it makes sense to do it for this one, too, I think -- rather than come up with a new approach. Thoughts?

anthony-tuininga:
I've just pushed code that resolves the pytz issue using the environment variable I mentioned. It seems simple and straightforward. If you agree, please close this pull request. If not, let me know and we can discuss further. Thanks!

marcelotduarte:
I know this method. The objective is to hide it in library.zip as pytz can load the data using pkg_resources intenally. In the sample I use zip_include_packages=["*"].
If you use default zip_include_packages empty, zoneinfo is copied (and visible).

marcelotduarte:
And with your patch I think that you'll force me to use it, and distribute thousand of files.

anthony-tuininga:
I am going to assume that means you disagree. :-) I understand your desire to not distribute all of the time zone files -- and yet that is exactly what the Python distributions do. Someone did try to place all of the modules in a zip file and actually found the performance of Python to suffer when that is done -- so there is always a trade-off between performance and ease of distribution.

One other issue, though -- if you are not including the pytz module in a zip file you definitely don't want to include the timezone files in the zip file. I'm not sure your patch accounts for that? Can you check? I'll give this some more thought and get back to you on this issue.

marcelotduarte:
Yes, I disagree. I prefer without none patch, that what you do. With your patch, like I say before, force me to distribute many files. Before this patch I had the option to include it setting a environ inside the init script or include in zip file like others do (previosly I used py2exe and inserted the zoneinfo into the generated executable).
I know about performance, If I want performance, I use the default behavior of cx-freeze (zip_include_packages empty). I my cases, pytz is called sometimes, not all the time.
The other issue I will check...

marcelotduarte:
Tested with:
1) 'zip_exclude_packages': ["pytz"]
2) 'zip_exclude_packages': ["*"] and 'zip_include_packages': []
In both cases the zoneinfo is duplicated with my patch (in filesystem and in library.zip
With my inital setup it is only in library.zip, and the tests pass.
How I can detect at hook.py, in load_pytz that zip_exclude_packages is on?

anthony-tuininga:
I am working on a change which should make this work for you -- which I hope to commit tomorrrow!

marcelotduarte:
Ok. Can you revert the your last patch about pytz. I have tested with cx-freeze 6.0 and default zip_include_packages and its work correctly, because at freeze.py all necessary files are copied.

marcelotduarte:
See new issue #551, I think to implement it to use in this case of pytz, but can be used for any other package.

anthony-tuininga:
@marcelotduarte, I have just pushed changes that allow you to determine if a module is going to be added into the file system (or not) as well as added support for adding files (or all of the files in a directory) to the library.zip file. This supersedes this pull request -- which I hope doesn't cause too much trouble for you. Please let me know if you have any comments about any of this.

marcelotduarte:
Oh, looks good. Thanks!

PR #483, opened 2019-07-12
Enable (partially) to work under non-ASCII path on Windows.

Suzumizaki:
Consider building under virtual environments made by venv which contains non-ASCII characters.

Especially, the default user name often contains that characters under non-English mode Windows. That means the full path of "Documents" or "Desktop" etc. possibly  be contain such characters.

In fact, due to using imagehlp.h functions, this is partial solution. For example, under en-US environment, this patch affects nothing. Another example, when using Japanese mode Windows and using characters that is *not* included cp932, this patch would help none.

But this patch would much better user experience, because at least they can use their native characters.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -16,6 +16,7 @@ import struct
 import sys
 import time
 import zipfile
+import traceback
 
 import cx_Freeze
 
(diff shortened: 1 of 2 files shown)


Suzumizaki:
... Sorry, I made something wrong. I'll re-fix later.

Suzumizaki:
... I finished fix to catch up master branch. Thank you.

marcelotduarte:
In some PR I solved problems with names in UTF-8 on Windows and Linux. And as this PR is very old, and the code is different, I will close it.
But, if you still have a problem with non-ASCII characters, please feel free to contact me.
Thanks!

PR #484, opened 2019-07-13
replace the extension module loader using importlib machinery

marcelotduarte:
replace the extension module loader using importlib machinery

In py37 extension modules that load other modules causes errors like
SystemError: <method 'load_module' of 'zipimport.zipimporter' objects> returned NULL without setting an error
In some cases I see ImportError too...
Using importlib machinery instead of deprecated imp modernizes the code and possibly will be compatible with py38
Tested using py36 and py37 on windows and with py37 on linux

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -436,10 +436,6 @@ class ModuleFinder(object):
         
         elif type == imp.C_EXTENSION:
             logging.debug("Adding module [%s] [C_EXTENSION]", name)
-            if parent is None:
-                # Our extension loader (see the freezer module) uses imp to
-                # load compiled extensions.
-                self.IncludeModule("imp")
 
         # If there's a custom hook for this module, run it.
         self._RunHook("load", module.name, module)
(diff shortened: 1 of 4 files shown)


marcelotduarte:
I submit a test to demonstrate the efficiency of the patch. Running setup, three sample files are downloaded, two are from aiohttp and one from gevent.
Before applying the patch, both aiohttp examples give an error, as I mentioned earlier, an error in zipimport. Gevent works since there was an earlier patch for that.
After the patch, the three examples work, that is, fix the problem with aiohttp and keep working with gevent.
In fact, I get the impression that it fixes some weird behaviors I had even using py36.

anthony-tuininga:
@marcelotduarte, thanks for this. I have just pushed changes that are similar to what you have done here with a few tweaks. As such, I will close this.

marcelotduarte:
I am grateful to contribute to a project that helps me a lot.
I wish you had applied my patch and made adjustments after applying it, as I have seen in other cases. Just today you did that by accepting the 493 patch and then immediately applying your patches.
It was a laborious patch to unravel the problems and solve them, otherwise I could not release my py37 project.
Another patch I proposed, #494, if accepted, I ask you to apply it and make any corrections, if any, about it.
And having accepted it, I believe a version can be released for testing as people are asking. Without it, on Linux will still have many problems. This month and next I intend to contribute to some more things that come up.

anthony-tuininga:
Thanks for letting me know. I'll keep that in mind. And thanks again for your help!

PR #485, opened 2019-07-13
use importlib machinery and  get rid of imp...

marcelotduarte:
I'll do it in a few steps.


anthony-tuininga:
Thanks!

marcelotduarte:
Thank you for your best work.

PR #487, opened 2019-07-13
more importlib use

marcelotduarte:
importlib now has decode source, that operates similar to existing code here, but we need to open the file only one time.


anthony-tuininga:
This one seems redundant now.

PR #488, opened 2019-07-13
use pkgutil.read_code instead of checking manually to avoid changes in...

marcelotduarte:
skip_bytes in each modification of the header of pyc


anthony-tuininga:
This one seems redundant now.

PR #492, opened 2019-07-18
In Windows, obtain only the dependent .exe or .dll files.

marcelotduarte:
Other types will not be verified. This speeds up, but also eliminates or reduces binderror.


anthony-tuininga:
Thanks!

PR #493, opened 2019-07-19
[Windows] Add defaults to include tkinter surely.

Suzumizaki:
With Python 3.7, the environment variables both TCL_LIBRARY and TK_LIBRARY are not set.

This fix will find tcl/tk libraries and dlls automatically, that would help many many many peoples who want to include tkinter with cx_Freeze, all over the internet :)

Tested on Python 3.7.4(x64) and 3.7.3(x86).


Suzumizaki:
Thank you!

Issue #460, opened 2019-02-26
cx_Freeze source does not support `setup.py develop`


TTimo:
Debian Buster - Python 3.7 - cx_Freeze github master:

Observed: I can run `python3 setup.py develop` in the cx_Freeze tree, but when I try to package a small test project, I get an error `cx_Freeze.freezer.ConfigError: no base named Console`

Expected result: support `setup.py develop` mode so it's easier to install/debug/test.

Thanks!

anthony-tuininga:
Is this still an issue with cx_Freeze 6 just released? I just tried it myself and it worked fine. This was on Fedora 30, not Debian, though!

marcelotduarte:
cx_Freeze 6.1 has just been released.
Assuming this has been resolved.
If you had issues please re-open.

PR #502, opened 2019-09-05
support to 'python setup.py develop'

marcelotduarte:
or 'pip install -e path/to/cx_Freeze'

Tested on Ubuntu
Fix issue #460


marcelotduarte:
You will need to add to .gitignore:

cx_Freeze/bases/

cx_Freeze/util.*.so

cx_Freeze/util*.pyd

PR #505, opened 2019-09-12
default py3 dependencies

marcelotduarte:


Berserker66:
Would this mean no VC Runtime has to be installed seperately and can safely be removed from setup scripts?

marcelotduarte:
Oops ... if this happens, I think I did more than I wanted to, but let's wait for what tuininga will say.

marcelotduarte:
PR #512 fix the python3.dll dependency with pyqy5 sample...

cx_Freeze/freezer.py
```diff
@@ -221,9 +221,10 @@ def _GetDefaultBinIncludes(self):
         """Return the file names of libraries which must be included for the
            frozen executable to work."""
         if sys.platform == "win32":
-            pythonDll = "python%s%s.dll" % sys.version_info[:2]
-            return [pythonDll, "gdiplus.dll", "mfc71.dll", "msvcp71.dll",
-                    "msvcr71.dll"]
+            pythonDlls = ["python%s.dll" % sys.version_info[0],
+                          "python%s%s.dll" % sys.version_info[:2],
+                          "vcruntime140.dll"]
+            return pythonDlls
```
anthony-tuininga, line 227: Why drop gdiplus.dll, mfc71.dll, msvcp71.dll and msvcr71.dll? Why not just add vcruntime140.dll? And how does this work with include_mscvcr? I'm not a regular Windows user so can you explain what the intent of this is? Thanks!

marcelotduarte: Py3 do not use this dlls.

Issue #37, opened 2013-07-26
ALLUSERS property in MSI


anthony-tuininga:
Originally reported by: **Anonymous**

----------------------------------------

It is currently not possible to install the MSI produced with bdist_msi without admin privileges.
Could you add this feature?

http://msdn.microsoft.com/en-us/library/windows/desktop/aa367559%28v=vs.85%29.aspx

----------------------------------------
- Bitbucket: https://bitbucket.org/anthony_tuininga/cx_freeze/issue/37

anthony-tuininga:
*Original comment by* **Thomas Kluyver (Bitbucket: [takluyver](https://bitbucket.org/takluyver), GitHub: [takluyver](https://github.com/takluyver))**:

----------------------------------------

Heads up: if you can work out how to add this yourself (the relevant code is in cx_Freeze/windist.py), it's probably quickest to fix it and make a pull request.

anthony-tuininga:
*Original comment by* **Mario Benito Ríos (Bitbucket: [Maberi](https://bitbucket.org/Maberi), GitHub: [Maberi](https://github.com/Maberi))**:

----------------------------------------

I've just made the pull request #72 to provide this feature.

marcelotduarte:
cx_Freeze 6.1 has just been released.
Assuming this has been resolved.

PR #507, opened 2019-09-12
enable the possibility to switch from per-user to per-machine install

RomainHivet:
Defaut behaviour has been changed. Defaut is per-user, and all_user=True can change the behaviour to per-machine.

The new behaviour is the one recommended by Microsoft:
https://docs.microsoft.com/en-us/windows/win32/msi/single-package-authoring

this solves #37

### DIFF
--- a/cx_Freeze/windist.py
+++ b/cx_Freeze/windist.py
@@ -244,7 +244,7 @@ class bdist_msi(distutils.command.bdist_msi.bdist_msi):
                 ('MaintenanceForm_Action', 'Repair'),
                 ('ALLUSERS', '2')]
 
-        if not self.all_users.upper() == 'TRUE':
+        if not self.all_users:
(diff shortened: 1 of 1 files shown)


cx_Freeze/windist.py
```diff
@@ -373,6 +376,7 @@ def initialize_options(self):
         self.environment_variables = None
         self.data = None
         self.install_icon = None
+        self.all_users = 'False'
```
anthony-tuininga: Can you turn this into a real boolean option? Use boolean_options = ["all-users"] on the bdist_msi class. Then you can also just say ``if not self.all_users``.

RomainHivet: I was not very familiar with bdist_msi class, and wasn't sure that would work. Anyway it is done now!

PR #511, opened 2019-09-14
use console_scripts entry_points

marcelotduarte:
now cxfreeze and cxfreeze-quickstart run on windows without postinstall


jschueller:
ping ?

anthony-tuininga:
Thanks!

Issue #504, opened 2019-09-11
6.0 and PyQt5 import error: missing DLL


VegarS:
I get an import error (missing DLL) with cx_Freeze 6 and PyQt5. Can be reproduced with the PyQt5 sample files:

![bilde](https://user-images.githubusercontent.com/1231463/64682042-0d44e780-d481-11e9-92c1-c61614ddcbd4.png)

Seems similar to #328, but the workaround there is not working.

Versions:
Python 3.7.3
PyQt5 5.13.0
cx_Freeze 6.0 (from pip/pypi)

jpeg13:
Maybe you can get a workaround by manually copying the missing DLL (presumably MSVCP140.dll from Microsoft Visual C++ 2015 Redistributable package) into the PyQt5 subdirectory inside the build directory, or into the build directory itself. I get a similar error message with PySide and cx_Freeze 5.1.1 or 6.0 (but not 6.0b1), see [my comment](https://github.com/anthony-tuininga/cx_Freeze/issues/278#issuecomment-392075260) to #278.

VegarS:
It seems like several DLLs are missing from build\exe.win-amd64-3.7\PyQt5.

To get it working I made a build with cx_Freeze 5.1.1 and copied over the extra DLLs that 5.1.1 included in that directory: python3.dll, Qt5Core.dll, Qt5Gui.dll, Qt5Network.dll, Qt5Widgets.dll and VCRUNTIME140.dll

marcelotduarte:
I tested the sample file on Linux and it worked. I will test on Windows later.

marcelotduarte:
I tested on Windows with py37-32. The workaround is to copy python3.dll to the directory where the executable was made. cx_Freeze copies python37.dll, but does not copy python3.dll. The vcruntime140.dll may also have to be copied when it no longer exists on the system. I will do a PR later.

VegarS:
I can confirm that copying python3.dll to the executable directory works.

Berserker66:
Same deal on PySide2, python3.dll is missing there as well

stephenmhall:
adding an include_files to the python3.dll solves you forgetting to copy it manually.

tamenol:
I have the same issue. I downgraded to python3.5.4 and problem solved. cx_freeze could successfully make an .msi and it can run without errors.

zkn365:
“copy **python3.dll** to the directory where the executable was made” can solve my problem.
Python 3.7.4
[...]

Issue #457, opened 2019-02-06
Error on running setup.py build


abrathore09:
File "C:\Users\user\AppData\Roaming\Python\Python36\site-packages\cx_Freeze\hooks.py", line 437, in _qt_implementation
    _qtcore = __import__(name, fromlist=['QtCore']).QtCore
ValueError: PyCapsule_GetPointer called with incorrect name

marcelotduarte:
cx_Freeze 6 has been released. Can you try again?

senyai:
Same:
```...
File "d:\python37_x86_64\lib\site-packages\cx_Freeze\finder.py", line 297, in _ImportModule
    deferredImports, namespace = namespace)
  File "d:\python37_x86_64\lib\site-packages\cx_Freeze\finder.py", line 390, in _InternalImportModule
    parentModule, namespace)
  File "d:\python37_x86_64\lib\site-packages\cx_Freeze\finder.py", line 443, in _LoadModule
    self._RunHook("load", module.name, module)
  File "d:\python37_x86_64\lib\site-packages\cx_Freeze\finder.py", line 510, in _RunHook
    method(self, *args)
  File "d:\python37_x86_64\lib\site-packages\cx_Freeze\hooks.py", line 557, in load_PyQt4_QtGui
    name, QtCore = _qt_implementation(module)
  File "d:\python37_x86_64\lib\site-packages\cx_Freeze\hooks.py", line 468, in _qt_implementation
    _qtcore = __import__(name, fromlist=['QtCore']).QtCore
ValueError: PyCapsule_GetPointer called with incorrect name```

marcelotduarte:
cx_Freeze 6.1 has just been released.
Assuming this has been resolved.
If you had issues please re-open.

Issue #455, opened 2019-01-26
possible bug with including vcruntime dll on windows


LakiWizard:
Greetings! I have been using cx_freeze to package a small game I made using pygame, and recently discovered that it may not work without python installed due to a missing file VCRUNTIME140.dll. The docs suggested to use the include_msvcr switch when building, which I added to the options in my setup.py file. However, cx_freeze seems to place the dll in the pygame subdirectory in the build directory, so I needed to manually move it where the executable is before it could work. This is on windows 10 with python 3.6.6 and cx_freeze 5.1.1

LakiWizard:
It seems to be the same problem as #372 except the dll is copied into lib/pygame rather than lib, for some reason.

marcelotduarte:
cx_Freeze 6 has been release. Can you try again? Can I close the issue?

marcelotduarte:
cx_Freeze 6.1 has just been released.
Assuming this has been resolved.
If you had issues please re-open.

Issue #372, opened 2018-04-05
Microsoft Visual C runtime DLLs are copied to the wrong folder


alessandrodd:
Could be related to #328 ; 
when using the option `'include_msvcr': True` , VCRUNTIME140.dll gets copied to the **lib** folder instead of its parent folder, i.e. the one that will contain the produced .exe:

![cattura](https://user-images.githubusercontent.com/11824406/38361188-daabd0be-38cc-11e8-9111-34850b434227.PNG)

causing the following error:

![err](https://user-images.githubusercontent.com/11824406/38361347-5613ad94-38cd-11e8-9499-577a4f74e41a.jpg)

Moving the VCRUNTIME140.dll from /lib to its parent directory fixes it.

FreeHarry:
I can confirm this issue for cx_freeze 6.0b1 (the newest one from GitHub) at Windows 10. Move the file `VCRUNTIME140.dll` to the parent folder seems to fix the error.

And on my machine (Win 10) it's equal if `include_msvcr` is `True or False`

FreeHarry:
What I also recognized is that the `python36.dll` is copied into both folders: exe-folder and lib-folder. If I remove the duplicated `python36.dll` from the lib-folder and move the `VCRUNTIME140.dll` to the exe-folder all runs well.

So it seems `python36.dll` and `VCRUNTIME140.dll` must be in the same folder as the executable and NOT in the lib-folder.

marcelotduarte:
cx_Freeze 6.1 has just been released.
Assuming this has been resolved.
If you had issues please re-open.

PR #512, opened 2019-09-14
fix issues with dll dependencies

marcelotduarte:
fix a regression introduced in #492
fix multiple get dependencies when the filename has different case

fix issues #504, #173 and maybe the related issues: #372, #328, #457, #455

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -254,7 +254,7 @@ class Freezer(object):
         dependentFiles = self.dependentFiles.get(path)
         if dependentFiles is None:
             if sys.platform == "win32":
-                if path.endswith(('.exe', '.dll')):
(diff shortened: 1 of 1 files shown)


marcelotduarte:
This fix issues #504, #173 and maybe the related issues: #372, #328, #457, #455

rerpha:
Are you planning on pushing a new release to pypy for this? I would like one as previously cx_freeze did not work for me on windows at all and it now does. Perhaps even just a minor/patch version? thanks. @anthony-tuininga

marcelotduarte:
https://github.com/rerpha
Anthony has little spare time. He intended to release a version the week after commit. Let's hope he gets this week.

rpodgorny:
it took me hours of trial and error and googling to finally land here. please make a stable release so others don't have to go through the same agony... ...thanks! ;-)

Issue #513, opened 2019-09-16
zipimport.ZipImportError on python3.6/encodings/__init__.py


GGui:
Hello,

I'm trying to execute the https://github.com/anthony-tuininga/cx_Freeze/tree/master/cx_Freeze/samples/advanced with docker.

I built my docker image everything is ok ("docker build -t test1 .")
But i'm trying to run my image,i get the following exeption:

"Fatal Python error: Py_Initialize: Unable to get the locale encoding
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/encodings/__init__.py", line 31, in <module>
zipimport.ZipImportError: can't decompress data; zlib not available"

I can't find a solution. I may have forgotten something.

Thank you.

My docker file is:
FROM python:3.6-alpine3.6 as builder

RUN apk --no-cache upgrade && apk --no-cache add build-base tar musl-utils openssl-dev && pip3 install --upgrade pip
RUN pip3 install setuptools cx_Freeze

COPY . .
RUN ln -s /lib/libc.musl-x86_64.so.1 ldd
RUN ln -s /lib /lib64
#RUN pip3 install -r requirements.txt
RUN python3 setup.py build_exe

FROM alpine:3.6
RUN apk --no-cache upgrade && apk --no-cache add wget zlib
COPY --from=builder build/exe.linux-x86_64-3.6 /test/
USER nobody:nobody
ENTRYPOINT ["/test/advanced_1"]

marcelotduarte:
In ubuntu this library is installed with the package 'zlib1g-dev', in alpine uses '-dev'?
I also think it has to be installed before you run python.
And I also think it's much more of a stackoverflow help question than a cx_Freeze issue.

GGui:
Sorry but i thought it was an issue.
I think that the issue is not due to the "zlib-dev" lib. I change in my docker file from zlib to zlib-dev and there is no change.

In the seconde part of my docker file, i create an alpine image without installing python to run the exe file builded in the first part of my docker file.

I thought weird that the exe file try to read /usr/local/lib/python3.6/encodings/init.py while my image don't contain "python", does it?

Is that normal behavior?

Thank you for your help.

marcelotduarte:
You mean two dockers right?
An executable generated by cx still has the same needs as python. It is a built-in python, where .py is compiled in bytecode. So something is probably missing from the docker to run.

GGui:
[...]

Issue #506, opened 2019-09-12
Problems freezing geospatial libraries


kurt-rhee:
Hello, 

I am trying to get around the sqlite dll error, so I use the following command from pip:

```
pip install --upgrade git+https://github.com/anthony-tuininga/cx_Freeze.git@master
```

I get the following error:
```
**error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": https://visualstudio.microsoft.com/downloads/**
```

I am having a hard time knowing which one is 14.0 as they are versioned by year instead of version number. 
https://visualstudio.microsoft.com/vs/older-downloads/

Thank you for your help.

marcelotduarte:
As I told you in issue #296, if you use `pip install -U cx-freeze`, you will be using the latest version. You do not need to compile. Try this first.

But if you still want to compile, VC 14.0 is known as VC 2015, and is used to compile python and hence extensions (https://docs.python.org/3.7/using/windows.html?highlight=visual#compiling-python-on-windows).

kurt-rhee:
Hey Marcelot, 

Thank you for being so responsive.  I have tried ```pip install cx-freeze``` and ```pip install -U cx-freeze```.  I also move the sqlite3.dll into the virtual environment manually.  The script completes with:

```
copying C:\Users\simon.rhee\AppData\Local\Continuum\anaconda3\envs\geo_env\DLLs\sqlite3.dll -> build\exe.win-amd64-3.6\lib\sqlite3.dll
```

And then finishes without any errors displayed.  

The resulting file structure has nothing inside of the dist folder.  Inside of the build folder there is an exe which runs, but cannot find fiona.ogrext import Iterator, ItemsIterator, KeysIterator

marcelotduarte:
So the problem is not what you think. The build was created, the exec was created. I recommend you read the documentation, some configuration was missing from your setup.py.
See: https://cx-freeze.readthedocs.io/en/latest/faq.html#problems-with-running-frozen-programs

kurt-rhee:
For people who may be finding this post in the future:

I was not able to fix the problem from the link above.  Loading the sqlite3 dll manually works fine, but finding the fiona.ogrext dll was a more challenging task.  I searched my virtual environment folders for dlls for both fiona and gdal.  I think I will try another methodology for packaging into an exe.

marcelotduarte:
[...]

PR #516, opened 2019-09-23
fix dependencies in a regression introduced by #297

marcelotduarte:
[Windows] some dlls are not copied (#506)
[Linux] ubuntu I didn't see any problems, but on alpine linux, has relations with this code (issue #513).


anthony-tuininga:
Thanks!

PR #517, opened 2019-09-23
alpine linux support

marcelotduarte:
For alpine linux to work fully in a docker, with no workarounds, other patches already shipped (#494, #516) are required and also one that I am finalizing. A sample can be done based on issue #513.


marcelotduarte:
#518

Issue #425, opened 2018-09-18
5.1.1, 6.0b1: build doesn't load libpython on linux (python 3.5, 3.6)


skozlovf:
Executable created with `5.1.1`, `6.0b1` and b18b4c3c fails with error:
```
Fatal Python error: Py_Initialize: Unable to get the locale encoding
Traceback (most recent call last):
  File "/home/vuser/.pyenv/versions/3.6.6/lib/python3.6/encodings/__init__.py", line 31, in <module>
zipimport.ZipImportError: can't decompress data; zlib not available
```

I used docker image with python `3.5.6`, `3.6.6` build with `pyenv`. The exe tested in various docker containers.

If run with `LD_LIBRARY_PATH` set then it runs correctly. So the issue is that the exe doesn't load `libpython3.X.so` because it searches in wrong paths (it doesn't search in the exe parent and `lib` dirs)

Some of the search paths (from `strace` log):
```
/usr/lib/x86_64-linux-gnu/tls/x86_64/libpython3.6m.so.1.0
/usr/lib/x86_64-linux-gnu/tls/libpython3.6m.so.1.0
/usr/lib/x86_64-linux-gnu/x86_64/libpython3.6m.so.1.0
/usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0
/lib/x86_64-linux-gnu/libpython3.6m.so.1.0
/usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0
/lib/libpython3.6m.so.1.0
/usr/lib/libpython3.6m.so.1.0
```

At the same time it correctly loads some other libs:
```
stat("/project/build/exe.linux-x86_64-3.6/lib/library.zip", {st_mode=S_IFREG|0644, st_size=807258, ...}) = 0
open("/project/build/exe.linux-x86_64-3.6/lib/library.zip", O_RDONLY) = 3

open("/project/build/exe.linux-x86_64-3.6/lib/zlib.cpython-36m-x86_64-linux-gnu.so", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\260\34\0\0\0\0\0\0"..., 832) = 832

```
Here is repo with demo scripts and `strace` logs: https://github.com/skozlovf/cxfreeze-tests.
Just run `./runtests.sh`. Take a look please.

PS probably related issues #324, #392.

**UPD1**:

[...]

PR #518, opened 2019-09-27
Always copy the python dynamic libraries into lib folder

marcelotduarte:
This patch has multiple purposes. With it libpython*.so* is moved to the lib folder, so only the executables will be in one folder and everything else in the lib folder. I see that in some cases, and there are several issues about it, libpython is copied several times. Now it's one less (To be honest, in linux I only saw duplicate, one in root and one in lib).
With this patch it was also possible to run alpine linux on a docker (with the help of #494 and #516). It will also resolve cases where the user compiles python from sources (in alpine it is the same). Why? Because python is placed in /usr/local and not in /usr, and ldd doesn't find the dependencies, so I find the python dependencies (sys.executable) and it resolves it.
I tested on ubuntu and alpine, with one of my projects and with samples advanced, sqlite, pyqt5.
I want to do the same for windows, but I have to solve the dll hell, and if it works, will greatly improve the issue of duplicate files.


marcelotduarte:
related issues #425, #513

cx_Freeze/freezer.py
```diff
@@ -215,7 +227,7 @@ def _GetDefaultBinExcludes(self):
         if sys.platform == "win32":
             return ["comctl32.dll", "oci.dll", "cx_Logging.pyd"]
         else:
-            return ["libclntsh.so", "libwtc9.so"]
+            return ["libclntsh.so", "libwtc9.so", "ldd"]
```
anthony-tuininga, line 230: What is the meaning of the addition of "ldd"? What would request its inclusion anyway?

setup.py
```diff
@@ -28,7 +28,7 @@ def build_extension(self, ext):
             return
         if sys.platform == "win32" and self.compiler.compiler_type == "mingw32":
             ext.sources.append("source/bases/manifest.rc")
-        os.environ["LD_RUN_PATH"] = "${ORIGIN}:${ORIGIN}/../lib"
+        os.environ["LD_RUN_PATH"] = "${ORIGIN}:${ORIGIN}/lib"
```
anthony-tuininga: The original code was intended so that if you installed the executable in /usr/bin (for example) that you would then find the library in /usr/lib. This would eliminate that and require that the files be found in /usr/bin/lib -- which seems wrong! Perhaps it might be satisfactory to simply add ${ORIGIN}/lib and not replace ${ORIGIN}/../lib"?

And if we are moving all files to the subdirectory lib, should we remove ${ORIGIN}?

cx_Freeze/freezer.py
marcelotduarte: On Alpine Linux ldd is a reference to a system library (virtual library maybe)

marcelotduarte: See #513 for more reference

marcelotduarte: Previously I had a patch #517 just for that, but I decided to put it together.

setup.py
marcelotduarte: I had been a little doubtful about this point, but some change here is needed. I tested it on ubuntu and alpine and it seemed correct. I didn't understand why it had this "$ {ORIGIN} /../ lib". And I think this "$ {ORIGIN}" could be removed yes. I'll try.

marcelotduarte: I changed it to:
os.environ["LD_RUN_PATH"] = "${ORIGIN}/../lib:${ORIGIN}/lib"

The test pass on alpine linux (with this patch and other that it depend #494)
The test pass on my projects.

marcelotduarte: PR updated.

PR #533, opened 2019-11-21
Fix installation if AlwaysInstallElevated policy is set.

### LINKED ISSUE #463: msi installer not installing into specified directory
Hi,

I have been using cx_Freeze to build executables on Windows 7, but thought I would try building an msi using build_msi. The msi file is created ok, but it always installs into the c:\ drive, ignoring the specified directory. I found an identical error here:

https://sourceforge.net/p/cx-freeze/mailman/message/31155211/

I tried the suggestions in this message, include setting initial_target_dir (on the build_msi options in the setup.py file) and install_dir (on the install options), but the same thing occurs. I have also tried cx_Freeze versions 5.1.1 and 6.01b.

Any thoughts on this? If you gave me a clue on where to look, I might be able to figure it out and submit a fix.

Cheers,
Michael

florianbehrens:
This should fix #463.

### DIFF
--- a/cx_Freeze/windist.py
+++ b/cx_Freeze/windist.py
@@ -44,6 +44,9 @@ class bdist_msi(distutils.command.bdist_msi.bdist_msi):
             msilib.add_data(self.db, "Directory", self.directories)
         if self.environment_variables:
             msilib.add_data(self.db, "Environment", self.environment_variables)
+        # This is needed in case the AlwaysInstallElevated policy is set.
+        # Otherwise installation will not end up in TARGETDIR.
+        msilib.add_data(self.db, "Property", [("SecureCustomProperties", "TARGETDIR")])
         msilib.add_data(self.db, 'CustomAction',
                 [("A_SET_TARGET_DIR", 256 + 51, "TARGETDIR",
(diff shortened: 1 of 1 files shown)


cx_Freeze/windist.py
```diff
@@ -44,6 +44,9 @@ def add_config(self, fullname):
             msilib.add_data(self.db, "Directory", self.directories)
         if self.environment_variables:
             msilib.add_data(self.db, "Environment", self.environment_variables)
+        # This is needed in case the AlwaysInstallElevated policy is set.
+        # Otherwise installation will not end up in TARGETDIR.
+        msilib.add_data(self.db, "Property", [("SecureCustomProperties", "TARGETDIR")])
```
anthony-tuininga, line 49: What happens if the AlwaysInstallElevated policy is *NOT* set? Does this cause a problem in that case? If not, I'd be happy to merge this. I don't use Windows regularly enough to know the answer or easily create a test.

florianbehrens:
No, it does not cause obvious problems. Note that the given property was also used by the MSI creation of CPython 3.4 (more info in issue #463).

Thanks for your quick response!

PR #537, opened 2019-12-11
pytz sample

marcelotduarte:
sample for PR #482

### DIFF
--- a/cx_Freeze/samples/pytz/test_pytz.py
+++ b/cx_Freeze/samples/pytz/test_pytz.py
@@ -1,13 +1,16 @@
+'''sample to show the datetime in RFC1123 (timezone is required)'''
 from datetime import datetime
-from tzlocal import get_localzone
+import pytz
+import tzlocal
 
 RFC1123 = '%a, %d %b %Y %H:%M:%S %z'
 
 try:
-    tz = get_localzone()
-except:
-    import pytz
+    tz = tzlocal.get_localzone()
+    print('Using your local timezone:', tz.zone)
+except pytz.exceptions.UnknownTimeZoneError as exc:
+    print('Detected your local timezone:', exc.args[0])
+    print("WARNING: fail to load pytz timezones, fallback to UTC")
     tz = pytz.utc
-    print("WARNING: fail to load pytz timezones")
 finally:
     print(datetime.now(tz).strftime(RFC1123))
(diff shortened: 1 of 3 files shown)


cx_Freeze/samples/pytz/test_pytz.py
```diff
@@ -0,0 +1,13 @@
...
+RFC1123 = '%a, %d %b %Y %H:%M:%S %z'
+
+try:
+    tz = get_localzone()
+except:
+    import pytz
+    tz = pytz.utc
+    print("WARNING: fail to load pytz timezones")
+finally:
+    print(datetime.now(tz).strftime(RFC1123))
```
anthony-tuininga, line 16: tzlocal.get_localzone() simply returns "local" when the timzone information cannot be loaded -- but it still "works". :-) My suggestion is to use something like this instead:

```
import datetime
import pytz
import tzlocal

RFC1123 = '%a, %d %b %Y %H:%M:%S %z'

localTZ = tzlocal.get_localzone()
print("Local TZ:", localTZ)

otherTZ = pytz.timezone("US/Eastern")
print("Other TZ:", otherTZ)

print("Local Time:", datetime.datetime.now(localTZ).strftime(RFC1123))
print("Other Time:", datetime.datetime.now(otherTZ).strftime(RFC1123))
```

This yields an exception when frozen. What do you think?

marcelotduarte: My sample is to prove that is failing to load pytz zoneinfo, not as example of use.

marcelotduarte: localTZ = tzlocal.get_localzone()
print("Local TZ:", localTZ)
raises a exception, "don't still works" without the my patch or your patch when frozen.

marcelotduarte: updated the sample to be more verbose

anthony-tuininga: I think the issue is that the implementation of tzlocal that I am using (or the fact that I am using Fedora 31) causes the script you provided to "work" for me whereas it doesn't work for you. But if my suggested script fails for both of us, why not use it or something like it?

marcelotduarte (commented):
If I understand, you prefer a exception to indicate that fail. If so, ok.
Out of curiosity, what versions of pytz and tzlocal you are using?
I'm using pytz 2019.3 and tzlocal 2.0.0

Issue #543, opened 2019-12-20
TypeError: an integer is required (got type bytes)


Berserker66:
Python 3.8.1 on Windows 10 64Bit, using as of posting this current git cx_Freeze

requirements file for the project:

```
appdirs>=1.4.3
argon2_cffi>=19.2.0
colorama>=0.4.3
cx-Freeze>=6.0
openpyxl>=3.0.2
python-dateutil>=2.8.1
screeninfo>=0.6.1
jinja2>=2.10.3
PyNaCl>=1.3.0
cython>=0.29.14
passlib>=1.7.2
psutil==5.6.7
pywin32>=227; sys_platform == 'win32'
PySide2==5.14.0
PyYAML>=5.2
colour>=0.1.5
holidays>=0.9.11
tableprint>=0.8.0
pyodbc>=4.0.27
json_tricks>=3.13.4
```

```
Outputting to: build\exe.win-amd64-3.8
running build_exe
Traceback (most recent call last):
  File "E:/ConParser/setup.py", line 110, in <module>
    cx_Freeze.setup(
  File "C:\Program Files\Python38\lib\site-packages\cx_Freeze\dist.py", line 348, in setup
    distutils.core.setup(**attrs)
  File "C:\Program Files\Python38\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Program Files\Python38\lib\distutils\dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "C:\Program Files\Python38\lib\distutils\dist.py", line 985, in run_command
    cmd_obj.run()
  File "C:\Program Files\Python38\lib\site-packages\cx_Freeze\dist.py", line 219, in run
    freezer.Freeze()
  File "C:\Program Files\Python38\lib\site-packages\cx_Freeze\freezer.py", line 641, in Freeze
    self.finder = self._GetModuleFinder()
  File "C:\Program Files\Python38\lib\site-packages\cx_Freeze\freezer.py", line 362, in _GetModuleFinder
    finder = cx_Freeze.ModuleFinder(self.includeFiles, self.excludes,
  File "C:\Program Files\Python38\lib\site-packages\cx_Freeze\finder.py", line 141, in __init__
    self._AddBaseModules()
  File "C:\Program Files\Python38\lib\site-packages\cx_Freeze\finder.py", line 152, in _AddBaseModules
    self.IncludeModule("traceback")
[...]

PR #545, opened 2019-12-24
PEP570 added "positional only arguments" in Python 3.8

marcelotduarte:
fix issue #543 
reference https://bugs.python.org/issue37032
modified sqlite sample to reproduce the issue


anthony-tuininga:
Thanks, Marcelo!

PR #548, opened 2019-12-31
revert commit bc81a9aac23a0d2a6dec8261f97dc72fd01b9eb7

marcelotduarte:
PYTZ_TZDATADIR is unecessary at default zip_include_packages.

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -449,14 +449,6 @@ def load_pythoncom(finder, module):
             copyDependentFiles = False)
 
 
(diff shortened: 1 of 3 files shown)


anthony-tuininga:
If you'd like, you can use the new finder.ZipIncludeFiles() for this hook -- but note that you'll need to handle the case (such as in Fedora) where the zoneinfo files aren't found inside the package itself (see the existing hook). I'm not sure the extra complexity is worth it -- but if you think so, feel free to create a new pull request!

Issue #560, opened 2020-01-05
cx_freeze 6.1 fails in cxfreeze command line


msilano:
When trying to run cx_freeze 6.1 on Windows 10 against any python file, it fails with the error: 
TypeError: __init__() missing 1 required positional argument: 'constantsModule'.  Stack trace below.

Thanks!

C:\Users\mike\Box\Inbox\boxbot>cxfreeze test.py
Traceback (most recent call last):
  File "c:\users\mike\appdata\local\programs\python\python38-32\lib\runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\mike\appdata\local\programs\python\python38-32\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\mike\AppData\Local\Programs\Python\Python38-32\Scripts\cxfreeze.exe\__main__.py", line 7, in <module>
  File "c:\users\mike\appdata\local\programs\python\python38-32\lib\site-packages\cx_Freeze\main.py", line 165, in main
    freezer = cx_Freeze.Freezer(executables,
TypeError: __init__() missing 1 required positional argument: 'constantsModule'

C:\Users\mike\Box\Inbox\boxbot>type test.py
print("Hello")

marcelotduarte:
Thanks for reporting. PR #561 will solve this.
As a workaround, create a minimal setup.py manually or using the command line "cxfreeze-quickstart".

msilano:
Thanks @marcelotduarte.  Really appreciate the help.  
I tried out the change; introduced a slightly different error downstream on a Mac.  Will try on windows VM once it finishes updating itself.......

  File "/Library/Frameworks/Python.framework/Versions/3.8/bin/cxfreeze", line 5, in <module>
    main()
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/cx_Freeze/main.py", line 168, in main
    freezer = cx_Freeze.Freezer(executables, constants_module,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/cx_Freeze/freezer.py", line 97, in __init__
    self.constantsModules = list(constantsModules)
TypeError: 'ConstantsModule' object is not iterable

msilano:
Thanks for the quick fix.  Really appreciate the help.

ghost:
Hello i've the same issue (v 6.1). Is there a fix ?

[...]

PR #561, opened 2020-01-05
main - required positional argument constantsModule

marcelotduarte:
issue #560 
introduced by commit 92a6f177a73f31a0bd4408096521a6d4d03b8949

### DIFF
--- a/cx_Freeze/main.py
+++ b/cx_Freeze/main.py
@@ -157,12 +157,13 @@ def ParseCommandLine():
 def main():
     options = ParseCommandLine()
     executables = [cx_Freeze.Executable(options.script,
-                                        initScript = options.initScript,
-                                        base = options.baseName,
-                                        icon = options.icon,
-                                        targetName = options.targetName)
+                                        options.initScript,
+                                        options.baseName,
+                                        options.targetName,
+                                        options.icon)
                   ]
-    freezer = cx_Freeze.Freezer(executables,
+    constants_module = cx_Freeze.ConstantsModule()
+    freezer = cx_Freeze.Freezer(executables, constants_module,
             includes = options.includeModules,
             excludes = options.excludeModules,
             replacePaths = options.replacePaths,
@@ -173,4 +174,3 @@ def main():
             zipIncludes = options.zipIncludes,
             silent = options.silent)
     freezer.Freeze()
-


cx_Freeze/main.py
```diff
@@ -157,12 +157,13 @@ def ParseCommandLine():
...
     options = ParseCommandLine()
     executables = [cx_Freeze.Executable(options.script,
-                                        initScript = options.initScript,
-                                        base = options.baseName,
-                                        icon = options.icon,
-                                        targetName = options.targetName)
+                                        options.initScript,
+                                        options.baseName,
+                                        options.targetName,
+                                        options.icon)
```
anthony-tuininga, line 163: Why these changes? They don't have anything to do with the bug, do they? Can you remove them?

anthony-tuininga:
I've just pushed a fix that resolves this issue in a cleaner way, I believe. So I'll close this. Thanks for the suggestion, however!

varadhbhatnagar:
@anthony-tuininga Issue still comes when I install cx-Freeze from PyPI. I am using Ubuntu 18.04. Had to go back to 6.0.

marcelotduarte:
@varadhbhatnagar
Go back to 6.0 or go ahead:
pip install -U git+https://github.com/anthony-tuininga/cx_Freeze.git@master

cx_Freeze/main.py
MarkoShiva: Is there some special reason for not merging this commit? I have problem of fetching from pypi in the CI/CD environment. locally I patched the same file and it works. The problem is that on PyPi I don't see package updated and there is no development version available in pypi either so I can't rely on CI/CD to buld my executable for some projects. 
At least in my project nothing is broken by removing kwargs and making them args.

anthony-tuininga: Can you explain why switching from keyword args to positional args is an improvement? It may work just fine, but what is the advantage to making the change? If you can explain that and it makes sense to me I'll be happy to merge that commit!

MarkoShiva: Well I am certain that having a keyword arguments is nicer as you do not need to pay attention that much on names of parameters and their order. 
But however you had a grave bug in important project. You could either push update faster or accept marcelo commit.

Its sort of a big problem when Pypi start hosting projects as packages that fail. There should be checking for that automated for every working supported version of python. Imagine when I needed to edit a python package in venv to have it work with 3.8 and even then the CI/CD couldn't do that so I needed to downgrade.

I mean project is nice but its not good to leave people downloading your package from pypi that do not work for over a full month. I mean I still after more then a month of that issue have the same issue if I try to install normally 6.1 from PyPi. Will package there ever be updated?

PR #565, opened 2020-01-20
MSYS2/mingw libffi

### LINKED ISSUE #596: libffi-7.dll not found with msys2/mingw64
An error log can be found here: https://ci.appveyor.com/project/behrisch/texttest/builds/31035619/job/pkh1ru8vptyn039k
This is probably a consequence of the fix for #523 in #556. In the light of https://github.com/conan-io/conan/issues/2638 it is not clear to me how a workaround would look like.

jeromerobert:


behrisch:
This patch probably has an error. As far as I can see dll_name is not defined in https://github.com/anthony-tuininga/cx_Freeze/pull/565/files#diff-efd19075f749a32c9452d03876158d1fR116 and should be replaced by something like `os.path.basename(dll_path)`.

jeromerobert:
@behrisch fixed. The search folder was also wrong (`lib` instead of `bin`).  I missed that error because, for one reason libffi-6.dll was still copied (may be by an other hook). I'm not even sure that this hook is needed with Mingw64.

marcelotduarte:
> I'm not even sure that this hook is needed with Mingw64.

For msys2 it is not needed. The libffi-6.dll is detected and copied.

jeromerobert:
> For msys2 it is not needed. The libffi-6.dll is detected and copied.

Indeed. So here is a simpler fixe.

marcelotduarte (approved):

PR #570, opened 2020-01-28
PyQt5 hook -- fix sip import

cainesi:
The QtCore hook is failing for me for PyQt5.11 and later.  The issue seems to be that starting with 5.11 sip was included in the PyQt5 namespace rather than as a separate module.  Added some code to import "PyQt5.sip" (instead of "sip") starting from PyQt5.11.


marcelotduarte (commented):
I did a basic test, using the sample in the cx_Freeze, I didn't see the need for this patch.
With cx-freeze 6.1, PyQt5 5.14.1, PyQt5-sip 12.7.1
On linux - 2452 files/folders, 199.8MB (on exe build has PyQt5app, lib folder)
On WIndows, 2.278 files / 146 folders, 137MB
With your PR, same PyQt5/sip versions
On Linux - 2495 files/folder, 325,7MB (on exe build has PyQt5app, lib folder, plus 2 more folders, lot of files duplicated)
On windows - I gave up testing

cainesi:
Sorry, not sure exactly what you're testing with.  I've attached a very simple example to make sure we are on the same page. 

Tested this on windows with PyQt5 5.14.1 and PyQt5-zip 12.7.0 installed.  

When this is run without my PR there is 1 folder in the build directory ("lib").  When it is run with my PR there are two extra directories ("imageformats" and "platforms").  Those two extra directories are Qt plugins that are supposed to be being included by PyQt-related hooks.  Without the PR I think what is happening is that the hook to include the sip module (load_PyQt4_QtCore()) fails with an ImportError, which causes the subsequent hooks not to run (it also causes cx_Freeze to list the PyQt5.QtWidgets module as missing in the output from the build process).

Before PyQt5.14 it looks like PyQt could limp along without the platforms module, but now it just fails, which is how I noticed this. 

[example.zip](https://github.com/anthony-tuininga/cx_Freeze/files/4142778/example.zip)

marcelotduarte:
> When it is run with my PR there are two extra directories ("imageformats" and "platforms").

I see these two folders on linux too, and at lib/PyQt5/Qt/plugins. Because of this, there are duplicates.

Please test with sample in cx_Freeze repo, and I'll test with your example.
https://github.com/anthony-tuininga/cx_Freeze/tree/master/cx_Freeze/samples/PyQt5

cainesi:
Sorry, hadn't noticed the samples under the cx_Freeze directory.

Tried that, same results.  When running without the fix, the output of the build command includes:

> Missing modules:
> PyQt5.QtWidgets imported from pyqt5app__main__

which seems to be due (when I debug it) to the hook in load_PyQt4_QtCore() failing.

The resulting application still runs because the entire PyQt5 module (including sip and all the plugins) has been copied into the build directory.  But if PyQt5 is specified in zip_include_packages, then the application fails without the fix (because it cannot find sip).

So it seems like the bottom line here is that:
1) Without the fix some files are not being included because a glitch in the code is preventing some hooks from firing.
2) This does not actually matter since the whole PyQt directory is copied into the application *unless PyQt is included in the zip file*.
3) But if PyQt is included in the zip file, then things break without the fix.

Maybe a further fix would be to revise the PyQt-related hooks so that importing sip and the various calls to copy_qt_plugins() are conditional on PyQt being included in the zip file?

marcelotduarte:
> Sorry, hadn't noticed the samples under the cx_Freeze directory.
Tried that, same results. 

For me, it runs. I see the missing modules but it runs.
If I run it with: python setup.py build_exe --zip-include-packages=PyQt5
It doesn't runs, get a ModuleNotFound

I see that your example is using zip_include_packages. You have to fix this case without break the other. I have been making some modifications in order to keep a minimum of files out of lib. I think you have to do that too, if possible of course.

> Maybe a further fix would be to revise the PyQt-related hooks so that importing sip and the various calls to copy_qt_plugins() are conditional on PyQt being included in the zip file?

Yes, Anthony recently created a way to check if you are using fs or zip, this will make your case easier. See load_pytz for how you can use module.WillBeStoredInFileSystem() and others.

anthony-tuininga:
I just tried your example without any hooks whatsoever (I am reworking these) and it works just fine. As soon as I include the package in the zip file it fails with "No module named PyQt5.sip". Do you want it included in the zip file? Or is including it in the file system adequate?

cainesi:
> I see that your example is using zip_include_packages. You have to fix this case without break the other. I have been making some modifications in order to keep a minimum of files out of lib. I think you have to do that too, if possible of course.

I don't think the PR "breaks" anything.  The break is in load_PyQt4_QtCore(), which always fails with the current version of PyQt.  It's just sort of dumb luck that the result happens to omit some files that may not be needed anyway (if PyQt5 is not in the zip file). :)

I went back and checked what happened when PyQt5.10 was installed.  In that case cx-freeze (without the PR) included the extra files.

> Yes, Anthony recently created a way to check if you are using fs or zip, this will make your case easier. See load_pytz for how you can use module.WillBeStoredInFileSystem() and others.

Thanks, I will take a look at that.

cainesi:
> I just tried your example without any hooks whatsoever (I am reworking these) and it works just fine. As soon as I include the package in the zip file it fails with "No module named PyQt5.sip". Do you want it included in the zip file? Or is including it in the file system adequate?

Yes, I do not think the hooks are needed if PyQt5 is included in the file system.  I just added another commit to my branch to disable the hooks if the PyQt module is being included in the file system (but you probably have a better way of doing this).

It looks like PyQt5.sip consists of a .pyd file---if I am understanding things right, that can't go in the zip file either way?

PR #572, opened 2020-01-30
PEP8 compliant; py35+ compliant; pylint - dist module

marcelotduarte:
Do you accept patches like this?
You make reference 'to be PEP 8 compliant' in https://github.com/anthony-tuininga/cx_Freeze/issues/551#issuecomment-570829404
If you accept this, I'll make more.

### DIFF
--- a/cx_Freeze/dist.py
+++ b/cx_Freeze/dist.py
@@ -14,8 +14,8 @@ import sys
 import cx_Freeze
 from cx_Freeze.common import normalize_to_list
 
-__all__ = [ "bdist_rpm", "build", "build_exe", "install", "install_exe",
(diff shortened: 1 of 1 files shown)


anthony-tuininga:
Yes, I will accept them, but please hold off a bit until the hook changes I have in mind land. :-) I'll work around this particular one.

marcelotduarte:
I had chosen this module because I thought it would not interfere much with others and it was easier, just to test if you would accept. I think of doing others: main, setupwriter and windist. I won't change hooks, finder and freeze for now, okay?

anthony-tuininga:
Yes, that's fine. Thanks.

Issue #579, opened 2020-02-05
No module named 'pkg_resources.py2_warn' with setuptools 45.0.0


jpeg13:
After building an application depending on `pkg_resources`, the executable fails with a `ModuleNotFoundError: No module named 'pkg_resources.py2_warn'` when using setuptools 45.0.0. 

This can be reproduced e.g. by adding an `import pkg_resources` to the [simple sample].(https://github.com/anthony-tuininga/cx_Freeze/blob/master/cx_Freeze/samples/simple/hello.py)

Traceback:

    C:\JP\80_Own_tools\tests\setuptools_45.0.0>build\exe.win-amd64-3.6\hello.exe
    Traceback (most recent call last):
      File "c:\jp\98_bitbucket_sandboxes\github\cx_freeze\cx_Freeze\initscripts\__startup__.py", line 40, in run
        module.run()
      File "c:\jp\98_bitbucket_sandboxes\github\cx_freeze\cx_Freeze\initscripts\Console.py", line 37, in run
        exec(code, {'__name__': '__main__'})
      File "hello.py", line 7, in <module>
        import pkg_resources
      File "C:\JP\80_Own_tools\tests\setuptools_45.0.0\venv\lib\site-packages\pkg_resources\__init__.py", line 86, in <module>
        __import__('pkg_resources.py2_warn')
    ModuleNotFoundError: No module named 'pkg_resources.py2_warn'

This works just fin with setuptools 44.0.0.

jeromerobert:
For the record, already submitted as #565

jpeg13:
> 
> 
> For the record, already submitted as #565

I missed that, sorry, I looked only at the issues. At least we come with similar solutions!

jpeg13:
PR https://github.com/anthony-tuininga/cx_Freeze/pull/584 solves the issue.

PR #580, opened 2020-02-05
Update hook for pkg_resources for setuptools 45.0.0.

jpeg13:
- Fixes ModuleNotFoundError: No module named 'pkg_resources.py2_warn' with setuptools 45.0.0 (#579), see also https://github.com/pypa/setuptools/issues/1963
- Tested with setuptools 45.0.0 and 44.0.0 (backward compatibility)

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -423,8 +423,13 @@ def load_PIL(finder, module):
 
 
 def load_pkg_resources(finder, module):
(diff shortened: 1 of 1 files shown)


anthony-tuininga:
Thanks for the suggestion. @marcelotduarte has already created a pull request that covers this as well as further potential issues. Let us know if what he has done does not cover your situation, however!

PR #592, opened 2020-02-16
closes #591

### LINKED ISSUE #591: load_ssl and load__ctypes hooks for copying SSL and other DLLs don't work in a pipenv virtualenv
This is like issue #437, except that I'm running setup.py in a virtualenv created by pipenv.
In that case, the SSL DLLs need to be searched in `sys.path`, not just in `sys.base_prefix/"DLLs"`, because `sys.base_prefix` is something like `C:\Users\USERNAME\.virtualenvs\PROJECTNAME-HASH` but there is no `DLLs` subdirectory there.

Workaround in setup.py:

```
for dll_mask in ["libcrypto-*.dll", "libssl-*.dll"]:
    for dir_name in sys.path:
        found_dlls = glob.glob(os.path.join(dir_name, dll_mask))
        if found_dlls:
            dataFiles.append(sorted(found_dlls)[-1])
            print(f"... {dataFiles[-1]} found ...")
            break
    else:
        print(f"!!! {dll_mask} not found !!!")
```

vashek:


vashek:
Superseded by #619, closing.

PR #599, opened 2020-02-27
change report to not show missing modules that the user excluded as p…

marcelotduarte:
…ackages/modules or that are excluded by cx-freeze

I'll put some PR to hooks that can depend on this patch.

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -607,6 +607,11 @@ class ModuleFinder(object):
         self.excludes[name] = None
         self._modules[name] = None
 
+    def ExcludePackage(self, name):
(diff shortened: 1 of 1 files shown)


marcelotduarte:
@anthony-tuininga In another commit you say:
> I wanted to think about that a bit yet. ExcludePackage() is not really excluding a package. It is removing the name from the list of "bad" modules that are discovered. The global names in a module are intended to do similar things. ExcludeModule() will do it as well -- but you added some stuff for wildcards. So I think we want a different name...but I'm not sure what that is, yet! Thoughts welcome!

Related to 'but you added some stuff for wildcards', it is not really wildcard, I used "*", but it can be "p" (to indicated packages). I do an value different from None, as None indicates modules.

PR #606, opened 2020-02-27
support python 3.5.2 and up

marcelotduarte:
use dis._unpack_opargs

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -31,25 +31,6 @@ STORE_OPS = (STORE_NAME, STORE_GLOBAL)
 
 __all__ = [ "Module", "ModuleFinder" ]
 
-if getattr(dis, '_unpack_opargs', None) is None:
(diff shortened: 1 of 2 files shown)


anthony-tuininga:
Can you provide a bit more detail on why this patch is needed?

marcelotduarte:
Only to organize, to not test the python version in each call.
But for me, I think that the preferable is to remove and support py 3.5.2 and up because this code is valid only for py35. For py36 and up the unpack is different.

anthony-tuininga:
I agree. Supporting Python 3.5.2 and up is preferable to adding code for 3.5.0 and 3.5.1. So does that mean you'll close this and create a new PR that simply checks for 3.5.2 in setup.py?

marcelotduarte:
I'll update this PR.

marcelotduarte:
updated.

anthony-tuininga:
Yep. Thanks!

PR #608, opened 2020-03-01
support metadata to use by pkg_resources

marcelotduarte:
While freezing an application using pikepdf I found the need to add the dist-info directories (I decided not to use pkg_resources to determine them, and get them directly from the file system).

Edit:
This fixes #112, #216 and #586 
Edit2:
dependency removed, can be merged as is
Edit3:
Tested by a user in closed issue: https://github.com/anthony-tuininga/cx_Freeze/issues/438#issuecomment-621252979

PR #616, opened 2020-03-22
Add hook to importlib_metadata load.

lexa:
importlib_metadata is a library which provides an API for accessing an
installed package’s metadata (see PEP 566), such as its entry points or
its top-level name. It works by finding *.egg or *.dist-info files from
installed packages. Because cx_freeze discards *.egg and *.dist-info
files on freezing, it breaks importlib_metadata.

Even worse, on import the package tries to extract its own metadata and
fails. Which makes it unloadable.

As a workaround, include .dist-info file of the package itself, allowing
it to be loaded. It won't function properly anyway because it can't find
any other package.


marcelotduarte:
PR #608 includes metadata. I think that after it, this will be unnecessary.

lexa:
I agree this will not be necessary.

PR #633, opened 2020-04-16
Make needed dirs when using include_resources

QuantumEntangledAndy:
When using include_resources of bdist_mac it will not copy over any files to subfolders. This is because the destination sub folders are not created.

This patch addresses this.

### DIFF
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -47,7 +47,7 @@ class bdist_dmg(Command):
                 raise OSError('creation of Applications shortcut failed')
 
             createargs.append('-srcfolder')
-            createargs.append(self.buildDir + '/Applications')
+            createargs.append(os.path.join(self.buildDir, 'Applications'))
 
         # Create the dmg
         if os.spawnvp(os.P_WAIT, 'hdiutil', createargs) != 0:
@@ -191,7 +191,7 @@ class bdist_mac(Command):
 
             # let the file itself know its place
             subprocess.call(('install_name_tool', '-id',
-                             '@executable_path/' + fileName, filePath))
+                             os.path.join('@executable_path', fileName), filePath))
 
             # find the references: call otool -L on the file
             otool = subprocess.Popen(('otool', '-L', filePath),
@@ -235,7 +235,7 @@ class bdist_mac(Command):
                 # see if we provide the referenced file;
                 # if so, change the reference
                 if name in files:
-                    newReference = '@executable_path/' + name
+                    newReference = os.path.join('@executable_path', name)
                     subprocess.call(('install_name_tool', '-change',
                                     referencedFile, newReference, filePath))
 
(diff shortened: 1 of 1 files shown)


cx_Freeze/macdist.py
```diff
@@ -324,6 +324,10 @@ def run(self):
                 self.copy_tree(resource, self.resourcesDir + '/' +
                                destination)
             else:
+                parent_dirs = os.path.dirname(self.resourcesDir + '/' +
+                                              destination)
+                if not os.path.exists(parent_dirs):
+                    os.makedirs(parent_dirs)
```
marcelotduarte: You can use os.makedir(..., exist_ok=True)

marcelotduarte (changes requested):

cx_Freeze/macdist.py
QuantumEntangledAndy: Sure can do. Should mention that its a python 3.2+ option

marcelotduarte: cx-freeze is only py35+

QuantumEntangledAndy: That's good :)

QuantumEntangledAndy:
Done, anything else?

marcelotduarte (approved):

cx_Freeze/macdist.py
```diff
@@ -324,6 +324,9 @@ def run(self):
                 self.copy_tree(resource, self.resourcesDir + '/' +
                                destination)
             else:
+                parent_dirs = os.path.dirname(self.resourcesDir + '/' +
```
anthony-tuininga: Any reason you didn't use os.path.join(self.resourcesDir, destination)?

marcelotduarte: I think he kept the previous code. See, before and after it has similar code.

anthony-tuininga: Yeah. I see that. So should the other code be changed as well? There are two other occurrences. I guess since this is macOS specific code that it isn't strictly necessary. :-)

QuantumEntangledAndy: Yes I maintained the current form. Would you like me to add a commit changing them all to `os.path.join`?

anthony-tuininga: May as well. Thanks. After that I'll merge it.

QuantumEntangledAndy: Ok done, anything else?

cx_Freeze/macdist.py
```diff
@@ -47,7 +47,7 @@ def buildDMG(self):
                 raise OSError('creation of Applications shortcut failed')
 
             createargs.append('-srcfolder')
-            createargs.append(self.buildDir + '/Applications')
+            createargs.append(os.path.join(self.buildDir, '/Applications'))
```
marcelotduarte: typo here and 2 more
os.path.join(self.buildDir, 'Applications')

QuantumEntangledAndy: This is weird my git diff and editor report this
```
diff --git a/cx_Freeze/macdist.py b/cx_Freeze/macdist.py
index aa48f0c..dae37b5 100644
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -47,7 +47,7 @@ class bdist_dmg(Command):
                 raise OSError('creation of Applications shortcut failed')

             createargs.append('-srcfolder')
-            createargs.append(self.buildDir + '/Applications')
+            createargs.append(os.path.join(self.buildDir, '/Applications'))
```

marcelotduarte: yes, but should without the '/' no?

QuantumEntangledAndy: Oh I misunderstood thanks

QuantumEntangledAndy: Ok that was embarrassing but I think I got them all :)

Issue #634, opened 2020-04-16
Code signing and numpy on macos


QuantumEntangledAndy:
When using the code signing option to create a .app on macos certain packages will fail. Notably numpy and scipy. This is because they contain a hidden folder `.dylibs` which contains some of the libraries used by numpy and scipy, and hidden folder are not allowed inside .apps.

Following error occurs during code signing:

```
build/MyApp-1.0.0.app: bundle format unrecognized, invalid, or unsuitable
In subcomponent: build/MyApp-1.0.0.app/Contents/MacOS/lib/numpy/.dylibs
```

To resolve this would require finding and moving the hidden files (those starting with a `.`) and also adjusting the associated library references.

In the case of numpy these library references are `@loader_path` which usually are skipped. So perhaps should address that too.

Example of skipped `@loader_path`:

```
skip unknown file b'@loader_path/libgcc_s.1.dylib'
skip unknown file b'@loader_path/../.dylibs/libgfortran.3.dylib
```

marcelotduarte:
The referenced PR was merged. I can close this issue?

QuantumEntangledAndy:
The issue is not directly related to that PR and is not resolved by it.

It's a bit more complex. Folders that start with a dot are forbidden when code signing. Some packages like numpy contain such folders.

To resolve there would have to be a means to rename the folders and update the reference in the library files.

marcelotduarte:
As I don't develop for mac, I may not have understood the problems. But, the PR #590 by @cainesi resolves this issue?
Can you test it?
`pip install -U git+https://github.com/marcelotduarte/cx_Freeze.git@refs/pull/590/head`

cainesi:
The PR might help. But I think you’d also need to mark the packages to be included in the zip file (that way all the binary dependencies would be dumped into one directory and you would not need to worry about the original package structure with the hidden folder). 

As a fringe benefit, your bundled app may get a lot smaller.

QuantumEntangledAndy:
My MacBook is currently in the repair shop but I'll test when it comes back.

I'm not familiar with the option to include packages in the zip. But the docs suggest it's a simple as adding the `zip_include_packages` option. If that's the one you mean I will try it when my Mac is back.

cainesi:
Yes, that’s the option I had in mind. 

Best,
Ian

marcelotduarte:
cx_Freeze 6.5 has just been released.
Assuming this has been resolved.
If you had issues please re-open.

PR #635, opened 2020-04-17
Check for Mach-O using byte strings to allow case of non unicode chars

### LINKED ISSUE #529: setRelativeReferencePaths() Error when trying to freeze wxPython app
Hi,

OS - MacOS Mojave
Python version: 3.7.5

When trying to freeze a wxPython application cx_Freeze fails with following traceback

`
File "../../Python.framework/Versions/3.7/lib/python3.7/site-packages/cx_Freeze/macdist.py", line 180, in setRelativeReferencePaths
if "Mach-O" in p.stdout.readline().decode():
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 281-282: invalid continuation byte
`

Any suggestions on how to proceed?

QuantumEntangledAndy:
The recent change of not encoding byte strings in bdist_mac is breaking the build. Methods using bytes should not be mixed with str. This patch unifies the relevant code to use all bytes.

NOTE: There is one step where I choose to `decode`

```python
else:
                        files.append(name.decode())
```

This is because 
```python 
filePath = os.path.join(self.binDir, fileName)
```

The `fileName` is generated from type `str` (from `os.walk`) but when we add extra files in the loop they were added as type bytes.

Currently I decode back to string then append to the list, but alternatively it may be better to encode the results of `os.walk` so that it is all bytes throughout.

### DIFF
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -177,7 +177,7 @@ class bdist_mac(Command):
         for root, dirs, dir_files in os.walk(self.binDir):
             for f in dir_files:
                 p = subprocess.Popen(("file", os.path.join(root, f)), stdout=subprocess.PIPE)
-                if "Mach-O" in p.stdout.readline().decode():
+                if b"Mach-O" in p.stdout.readline():
                     files.append(os.path.join(root, f).replace(self.binDir + "/", ""))
 
         for fileName in files:


QuantumEntangledAndy:
I have a further patch that gets `@loader_path` working (see #634) BUT when I came to merge I had conflicts from your changes to not decode. When I pulled your changes I ran into multiple errors related to using `str` when `bytes` was expected.

If you want I can roll the addition of `@loader_path` into this pull request too. Or I can wait until this is merged and pull request of that.

marcelotduarte:
That is my patch. It was untested and I forget to declare this. In the issue #529 I requested tests but no one answered. As you use mac, if you find it better to revert the patch and you fix the issue, I even prefer it.

QuantumEntangledAndy:
I actually had #529 and had a different fix. Rather than not decoding I did this instead
```
try:
    line = p.stdout.readline().decode()
except UnicodeDecodeError:
    continue
if "Mach-O" in line:
```

Here I assume that if we can't decode it is not Mach-O. Which fixes #529 in my case I am not sure what the original case was.

We have two choice

1. resolve by changing all to bytes (many changes, but possibly more flexible)

2. resolve by skipping any files where it is not decodable (less changes, but possibly not as flexible).

Which would you prefer?

marcelotduarte:
I reverted my patch. Please send your patch, your approach is ok to me, the option 2.

QuantumEntangledAndy:
I have made this pull request work as indicated in option 2.

Fixes #529

QuantumEntangledAndy:
Actually I am thinking of an option 3.

Use str everywhere except when testing for Mach-O this will let us test for Mach-O in files with non unicode bytes, while still not requiring many changes

marcelotduarte:
ok, put it in another PR

QuantumEntangledAndy:
A completely separate PR? Or ammend this one?

marcelotduarte:
Can be in this PR, but change the title to not confuse anthony.

QuantumEntangledAndy:
Latest commit is using option number 3. This allows searching for Mach-O even if there are non unicode chars but keeps the rest of the code as str

marcelotduarte (approved):

PR #636, opened 2020-04-17
Revert "decode error on mac"

marcelotduarte:
Reverts anthony-tuininga/cx_Freeze#573


marcelotduarte:
@anthony-tuininga the previous patch was untested. I'm reverting it. Another developer will fix it.

PR #637, opened 2020-04-17
Handle @loader_path references

QuantumEntangledAndy:
This patch allows for the handling of `@loader_path` libraries references.


QuantumEntangledAndy:
This works however I have a case in the package panda3d where resources are loaded relative to the location of the dylib.

Panda3d needs special handling (by the app using panda3d not necessarily by cx_freeze where the location of the config file is specified manually).

I was planning part of this to be applied to fix issues with code signing of numpy and scipy but perhaps this will be more trouble than it is worth and we should just leave `@loader_path`

PR #641, opened 2020-04-26
fix optimize option in python 3.8

### LINKED ISSUE #639: Assertions aren't removed when "optimize=2"
Tested on Windows 10, Python 3.8.2, freezing a WinGUI application.
Doesn't work on https://github.com/anthony-tuininga/cx_Freeze/commit/485eb0330de934a6418afa42b2eb82f1948cf2d7, https://github.com/anthony-tuininga/cx_Freeze/commit/485eb0330de934a6418afa42b2eb82f1948cf2d7 and 6.1 release.

Did work on https://github.com/anthony-tuininga/cx_Freeze/commit/7f0ceefa89ab48896bb28c8f57c0e550b338d9d8

Minimal failing test case:

test.py:
```
assert False
print("Hello")
```

setup.py:

```
from setuptools import find_packages

executable_options = dict(
    base="Win32GUI",
    targetName="test.exe"
)

build_exe_options = dict(
    packages=find_packages(),
    optimize=2
)

setup(
    name="Assert Test",
    version="1.0",
    options=dict(build_exe=build_exe_options),
    executables=[
        Executable("test.py", **executable_options)
    ]
)
```

marcelotduarte:
fix #639
I looked for something in the python documentation, indicating the change but I couldn't find it. I did some tests and saw that in py38 just passing the parameter -O or -OO globally set the optimize flag (it didn't work with the method that existed or using os.environ ['PYTHONOPTIMIZE']). So I decided to propagate the variable to the finder method and it is even easy to see.
So, only on win32 will we need 'cx_Freeze.util'.
Success: https://github.com/anthony-tuininga/cx_Freeze/issues/639#issuecomment-620165863

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -653,6 +653,9 @@ class ModuleFinder(object):
     def SetOptimizeFlag(self, optimizeFlag):
         """Set a new value of optimize flag and returns the previous value."""
         previous = self.optimizeFlag
+        # The value of optimizeFlag is propagated according to the user's
+        # choice and checked in dist.py or main,py. This value is unlikely
+        # to be wrong, yet we check and ignore any divergent value.
         if -1 <= optimizeFlag <= 2:
             self.optimizeFlag = optimizeFlag
         return previous
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -607,7 +607,7 @@ class ModuleFinder(object):
             moduleName = name
         info = (ext, "r", imp.PY_SOURCE)
         deferredImports = []
-        module = self._LoadModule(moduleName, 0, path, info, deferredImports)
+        module = self._LoadModule(moduleName, None, path, info, deferredImports)
         self._ImportDeferredImports(deferredImports)
         return module
 
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -125,6 +125,7 @@ class ModuleFinder(object):
             constants_module=None, zip_includes=None):
         self.include_files = include_files or []
         self.excludes = dict.fromkeys(excludes or [])
+        self.optimizeFlag = 0
         self.path = path or sys.path
         self.replace_paths = replace_paths or []
         self.zip_include_all_packages = zip_include_all_packages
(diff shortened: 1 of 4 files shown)


cx_Freeze/finder.py
```diff
@@ -605,8 +607,7 @@ def IncludeFile(self, path, moduleName = None):
             moduleName = name
         info = (ext, "r", imp.PY_SOURCE)
         deferredImports = []
-        module = self._LoadModule(moduleName, open(path, "U"), path, info,
-                deferredImports)
+        module = self._LoadModule(moduleName, 0, path, info, deferredImports)
```
anthony-tuininga: What's the reason for passing 0 instead of open(path, "U") for the file handle? Is this intended?

cx_Freeze/finder.py
```diff
@@ -649,6 +650,13 @@ def ReportMissingModules(self):
                              "may not be needed on this platform.\n")
             sys.stdout.write("\n")
 
+    def SetOptimizeFlag(self, optimizeFlag):
+        """Set a new value of optimize flag and returns the previous value."""
+        previous = self.optimizeFlag
+        if -1 <= optimizeFlag <= 2:
+            self.optimizeFlag = optimizeFlag
+        return previous
```
anthony-tuininga, line 661: Should an exception be raised for this? Or should it just be passed through and let Python raise the exception when it sees the invalid value?

cx_Freeze/finder.py
marcelotduarte: In _LoadModule, when the type is imp.PY_SOURCE like this case, the source in the path is always open, then this is leaked. While debugging the optimize, I see this. I'm looking ahead, in the py39 open with "U" is not permitted, the I removed this unnecessary code at this opportunity.

cx_Freeze/finder.py
marcelotduarte: As in general this value is already checked before reaching Freezer.py, and there it was not checked, so I just did a basic check. If it is a valid value, it is accepted, otherwise ignore it.

marcelotduarte: I added a comment.

cx_Freeze/finder.py
anthony-tuininga: So you're saying that argument is unused? For all supported Python versions? If so, perhaps it could be changed to `None`?

cx_Freeze/finder.py
anthony-tuininga: Ok.

cx_Freeze/finder.py
marcelotduarte: Yes, in _LoadModule, if the type is PY_SOURCE, it always opens the source.
Changed to none.

PR #642, opened 2020-04-27
better support for msys2 and anaconda3

### LINKED ISSUE #526: sqlite3.dll not always in DLLs folder
In Anaconda3 distributions, the sqlite3.dll is not in the `DLLs` folder, but in `Library/bin`.  As per issue #207, I modified hooks.py, load_sqlite3() as shown below to successfully load the dll from the Anaconda location.
```
def load_sqlite3(finder, module):
    """In Windows, the sqlite3 module requires an additional dll sqlite3.dll to
       be present in the build directory."""
    if sys.platform == "win32":
        dll_name = "sqlite3.dll"
        dll_path = os.path.join(sys.base_prefix, "DLLs", dll_name)
        lib_path = os.path.join(sys.base_prefix, "Library", "bin", dll_name)
        if os.path.exists(dll_path):
            finder.IncludeFiles(dll_path, os.path.join("lib", dll_name))
        else:
            finder.IncludeFiles(lib_path, os.path.join("lib", dll_name))
```

### LINKED ISSUE #566: tkinter path error
Hi,
Got a problem with version  mingw-w64-x86_64-python-cx_Freeze-6.0-1
Can't find path to tkinter (error: [Errno 2] No such file or directory: 'C:/msys64/mingw64/tcl/tcl8.6')
I try to add path in setup-freeze.py:
...
include_files = [
    r"C:\msys64\mingw64\bin\tcl86.dll",
    r"C:\msys64\mingw64\bin\tk86.dll",]

os.environ['TCL_LIBRARY'] = r'C:\msys64\mingw64\lib\tcl8.6'
os.environ['TK_LIBRARY'] = r'C:\msys64\mingw64\lib\tk8.6'
....
Got this error: error: [Errno 2] No such file or directory: 'C:/msys64/mingw64/Dlls/tcl86t.dll'
tcl86t.dll doesn't exist, my file is tcl86.dll

Thanks for help!

marcelotduarte:
msys2
- correct copy files of tkinter
- excludes DLLs to be copied as they are copied automatically and are in different folders than the python from python.org

anaconda3
- correct copy sqlite3.dll

fixes #566, #526 and others


cx_Freeze/samples/sqlite/test_sqlite3.py
```diff
@@ -29,4 +29,6 @@
         f.write('%s\n' % line)
 
 print('dump.sql created')
-con.close()
\ No newline at end of file
+con.close()
+
+input("Press ENTER")
```
anthony-tuininga: Why are you adding these lines in each of the samples?

marcelotduarte: To facilitate when testing on windows. In order not to have to open a command window and test it, because if you test directly in the window you are building with build\exe.....\test,exe, sometimes you pass the test, but outside, errors can occur when dll library is missing.

anthony-tuininga: Please remove those. They are annoying for those who aren't using your particular configuration. If you want you can always create a batch file that contains a "pause" in it. If you really want them for Windows, though, please add a check to make sure that you are on Windows -- so the rest of us don't have to deal with it. ;-)

marcelotduarte: ok. removed.

PR #647, opened 2020-05-02
cmd: Add --include-files option

xantares:
also remove the --ext-list-file option which was ignored

### DIFF
--- a/cx_Freeze/main.py
+++ b/cx_Freeze/main.py
@@ -94,11 +94,6 @@ def ParseCommandLine():
             dest = "excludeModules",
             metavar = "NAMES",
             help = "comma separated list of modules to exclude")
-    parser.add_option("--ext-list-file",
-            dest = "extListFile",
-            metavar = "NAME",
-            help = "name of file in which to place the list of dependent files "
-                   "which were copied into the target directory")
     parser.add_option("--include-files",
             dest = "includeFiles",
             metavar = "NAMES",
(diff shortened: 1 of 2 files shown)


xantares:
ping @anthony-tuininga @marcelotduarte

marcelotduarte:
@xantares, I believe that @anthony-tuininga will soon accept your patch.

cx_Freeze/main.py
```diff
@@ -94,11 +94,10 @@ def ParseCommandLine():
             dest = "excludeModules",
             metavar = "NAMES",
             help = "comma separated list of modules to exclude")
-    parser.add_option("--ext-list-file",
-            dest = "extListFile",
-            metavar = "NAME",
-            help = "name of file in which to place the list of dependent files "
-                   "which were copied into the target directory")
```
anthony-tuininga, line 101: Why are you removing the --ext-list-file option?

xantares: @anthony-tuininga  this option is unused: extListFile is not passed to Freezer, and I see no option in Freezer that could match its description, so I felt its best to remove it than silently ignore it

xantares: @anthony-tuininga are we good ?

anthony-tuininga: Yes. I see that indeed the option is unused.

Issue #536, opened 2019-12-06
Installing a Win32Service gives AttributeError: module '__main__' has no attribute 'NAME'


danielmenzel:
Hi,
I have built cx_Freeze 6.0 with Python 3.7.5, cx_Logging 2.2, Visual Studio 2019 on Windows 10 and have managed to get a `Win32Service.exe` in the `bases` directory.
I then built the provided sample under `samples/service`, but when I run the created executable with
```cmd
build\exe.win-amd64-3.7\cx_FreezeSampleService.exe --install test
```
I get the error
```
Service not installed. See log file for details.
```
and the following log file:
```
[20580] 2019/12/06 08:09:43.120 starting logging at level ERROR
[20580] 2019/12/06 08:09:43.134 Python exception encountered:
[20580] 2019/12/06 08:09:43.134     Internal Message: cannot locate service name
[20580] 2019/12/06 08:09:43.134     Type => <class 'AttributeError'>
[20580] 2019/12/06 08:09:43.135     Value => module '__main__' has no attribute 'NAME'
[20580] 2019/12/06 08:09:43.142     AttributeError: module '__main__' has no attribute 'NAME'

[20580] 2019/12/06 08:09:43.154 ending logging
```
It seems like the global variables in `Config.py` like `NAME` etc. are somehow not recognized as attributes of the `__main__` module.
I managed to get it working by explicitly adding the variables to the `__main__` module in `Config.py` like this
```python
import sys
sys.modules[__name__].NAME = 'cx_FreezeSampleService%s'
sys.modules[__name__].DISPLAY_NAME = 'cx_Freeze Sample Service - %s'
sys.modules[__name__].MODULE_NAME = 'ServiceHandler'
sys.modules[__name__].CLASS_NAME = 'Handler'
sys.modules[__name__].DESCRIPTION = 'Sample service description'
sys.modules[__name__].AUTO_START = False
sys.modules[__name__].SESSION_CHANGES = False
```
Perhaps there is something that can be changed in `Win32Service.c` instead of this inelegant workaround, but I have not enough knowledge about the Python C API to understand what is going on.

anthony-tuininga:
I don't use Windows much at all in the past few years but if I get a chance, I'll take a look at this. Its possible that some of the recent changes to simplify things have inadvertently broken this!

Matthias-Wagner:
[...]

PR #650, opened 2020-05-03
when building from sources, get cx_Logging to build Win32Service.exe

### LINKED ISSUE #519: Win32Service.exe base doesn't get built
When you build 6.0 the Win32Service.exe base doesn't get built, it's just silently ignored. You need to have cx_Logging available - not installed but a built distribution in a parallel directory without the version number (i.e. "cx_Logging" not "cx_Logging-2.2")

Also, the cx_Logging 2.2 tar ball is missing the cx_Logging.h so it doesn't build. I had to pull that from GitHub and drop it into cx_Logging\src

Finally, the cx_Freeze build will still fail if cx_Logging.h isn't also put into the root of cx_Logging!

Since cx_Logging hasn't changed in years could it not be included in the cx_Freeze tar ball so you can build Win32Service.exe without so much pain!?

marcelotduarte:
fix #519 
If the source cannot be obtained, it silently fails, as it currently does.
Also:
- Win32Service - replace deprecated functions that will be removed in py4
- cleanup samples

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,13 +88,27 @@ class build_ext(distutils.command.build_ext.build_ext):
 
 
 def find_cx_Logging():
-    dirName = os.path.dirname(os.getcwd())
+    import subprocess
+    dirName = os.path.dirname(os.path.dirname(os.path.os.path.abspath(__file__)))
     loggingDir = os.path.join(dirName, "cx_Logging")
+    if not os.path.exists(loggingDir):
+        try:
+            subprocess.run(["git", "clone",
+                            "https://github.com/anthony-tuininga/cx_Logging.git",
+                            loggingDir])
+        except (FileNotFoundError, subprocess.SubprocessError):
+            pass
     if not os.path.exists(loggingDir):
         return
     subDir = "implib.%s-%s" % (distutils.util.get_platform(), sys.version[:3])
     importLibraryDir = os.path.join(loggingDir, "build", subDir)
     includeDir = os.path.join(loggingDir, "src")
+    if not os.path.exists(importLibraryDir):
+        try:
+            subprocess.run([sys.executable, "setup.py", "install"],
+                           cwd=loggingDir)
+        except (FileNotFoundError, subprocess.SubprocessError):
+            pass
     if not os.path.exists(importLibraryDir):
         return
     return includeDir, importLibraryDir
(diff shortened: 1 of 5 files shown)


source/bases/Win32Service.c
```diff
@@ -494,10 +499,11 @@ static int Service_Uninstall(
         return LogWin32Error(GetLastError(), "cannot open service manager");
 
     // create service
-    serviceHandle = OpenServiceW(managerHandle, PyUnicode_AS_UNICODE(fullName),
-            SERVICE_ALL_ACCESS);
+    wfullName = PyUnicode_AsWideCharString(fullName, NULL);
+    serviceHandle = OpenServiceW(managerHandle, wfullName, SERVICE_ALL_ACCESS);
     if (!serviceHandle)
         return LogWin32Error(GetLastError(), "cannot open service");
+    PyMem_Free(wfullName);
```
anthony-tuininga: This should happen before the check on the service handle; otherwise, a potential memory leak occurs!

setup.py
```diff
@@ -88,13 +88,27 @@ def get_ext_filename(self, name):
 
 
 def find_cx_Logging():
-    dirName = os.path.dirname(os.getcwd())
+    import subprocess
```
anthony-tuininga, line 91: I think I'd prefer this to be a warning that cx_Logging was not found. The documentation (and the warning itself) can include the commands required to set things up, but they shouldn't just be run automatically, I think.

source/bases/Win32Service.c
marcelotduarte: I'll take a look.

setup.py
marcelotduarte: Only when compiling via sources will this happen, not in installations in general. For example, even if I had instructions, warning, or documentation I would prefer it to be automated and I think whoever opened the issue too. When I went to test issue #536, I had a good job to find out how to use cx_Logging, if it were automatic, it would be so simple.

source/bases/Win32Service.c
marcelotduarte: This leak and others resolved.

setup.py
anthony-tuininga: Ok. I can live with it. ;-)

PR #651, opened 2020-05-07
Replace the value of __package__ directly in the code

marcelotduarte:
This is an alternative to #609
Runs successfully with pike pdf and don't break samples with pyside2.

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -462,6 +462,9 @@ class ModuleFinder(object):
             # Scan the module code for import statements
             self._ScanCode(module.code, module, deferredImports)
         
+            # Verify __package__ in use
+            self._ReplacePackageInCode(module)
+
         module.in_import = False
         return module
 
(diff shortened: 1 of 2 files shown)


cx_Freeze/finder.py
```diff
@@ -480,6 +483,40 @@ def _LoadPackage(self, name, path, parent, deferredImports, namespace):
...
+            pkg_const_index = len(constants)
+            pkg_name_index = co.co_names.index("__package__")
+            if pkg_const_index > 255 or pkg_name_index > 255:
+                # Don't touch modules with many constants or names;
+                # This is good for now.
+                return
+            # The bytecode/wordcode
+            codes = [LOAD_CONST, pkg_const_index,
+                     STORE_NAME, pkg_name_index]
+            #if pkg_const_index > 255:
```
anthony-tuininga: Does this commented out code need to be present?

marcelotduarte: no.

marcelotduarte:
@anthony-tuininga this is ok?

anthony-tuininga:
I dont see anything obvious so I'm going to assume you've tested this with multiple Python versions and platforms. :-)

marcelotduarte:
Yes, tested in py35/32, 36/32, 37/32, 37/64, 38/64.
Currently, I need it only for pikepdf, but is a generic approach.

Issue #659, opened 2020-05-27
freezer._GetDependentFiles may be change `X_OK` to `R_OK` will be better


medns:
freezer.[_GetDependentFiles](https://github.com/anthony-tuininga/cx_Freeze/blob/c25fedd9c26a58139ac158259d35fc3d964fc509/cx_Freeze/freezer.py#L280)  May be change `X_OK` to `R_OK` will be better because a lot of dynamic library(*.so) no execute permission.

``` python
 if not os.access(path, os.X_OK):
```

change to:

``` python
 if not os.access(path, os.R_OK):
```

marcelotduarte:
Can you send a PR? Do you have a use case?

marcelotduarte:
Do you have a sample or use case?

marcelotduarte:
@medns Do you have a sample or use case?

PR #661, opened 2020-05-28
fix: dynamic library(*.so) is missing when there is no execute permission (#659)

medns:
fix: dynamic library(*.so) is missing when there is no execute permission (#659)


anthony-tuininga:
The problem with simply looking at any readable file is that for historical purposes shared libraries needed to have the executable bit set and a check was specifically added to avoid calling ldd and otool on files that aren't actually shared libraries -- in order to improve performance substantially! So I am reluctant to simply remove the check -- even though I recognize on modern systems this isn't strictly necessary. If you can resolve that issue that would be helpful.

PR #717, opened 2020-07-26
A sample for testing PyQt5 included in zip package.

cainesi:
Added a sample for testing that cx_Freeze works with PyQt5 included in the zip.


marcelotduarte:
Instead of creating a new sample, what do you think of creating just setup_zip.py?

cainesi:
Fine with me.  Whatever you prefer.

PR #718, opened 2020-07-26
Code to fix issues with PyQt5-related hooks. 

cainesi:
Code changes to let cx_Freeze locate sip on PyQt5.11 and later.

Also disabled certain hooks that are not required if the whole PyQt package directory is being copied into the package (rather than PyQ being included in the zip file).

This can be tested with the sample in PyQt5_zip (PR: #717).  This change is only helpful if/when the changes to finding OSX binary dependencies are accepted (PR: #590) (since without that, PyQt does not work from the zip file anyway).

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -744,6 +744,9 @@ def _qt_implementation(module):
               "Some incorrect files may be copied.")
     return name, _qtcore
 
+def _qt_in_file_system(module):
+    return module.WillBeStoredInFileSystem()
+
 def copy_qt_plugins(plugins, finder, QtCore):
     """Helper function to find and copy Qt plugins."""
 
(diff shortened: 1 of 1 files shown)


marcelotduarte:
@cainesi I merged this, but I forgot to ask you: Will it work on windows as well as on macOS?

cainesi:
> @cainesi I merged this, but I forgot to ask you: Will it work on windows as well as on macOS?

Yes. It was actually a problem with a project on windows that led me to make the change. I have not tested with Linux—but the basic issue is that the sip package moved, and I assume pyqt made the same change on all platforms.

Winand:
I think this prevents from calling `_QtGui` and platform plugins are not copied now. I use Anaconda (Windows 7 x64) and plugins are stored in a separate directory `Anaconda3\Library\plugins`

PR #728, opened 2020-07-28
Small fix to handle a build constant that includes a "=" symbol.

cainesi:
Small fix to deal with the case where a build constant includes an equal sign. For example:

```
var="abc=xyz"
```

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 import datetime
 import distutils.sysconfig
 from importlib.util import MAGIC_NUMBER
+from keyword import iskeyword
 import marshal
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


cx_Freeze/freezer.py
```diff
@@ -740,13 +747,16 @@ def __init__(self, releaseString = None, copyright = None,
...
+            parts = constant.split("=", maxsplit=1)
             if len(parts) == 1:
                 name = constant
                 value = None
             else:
                 name, stringValue = parts
                 value = eval(stringValue)
+            if not _isValidVariableName(name):
+                raise Exception("Invalid constant name in ConstantsModule (\"{}\")".format(name))
+
```
marcelotduarte: _isValidVariableName - for new funcions, you can use pep8 names. Anthony had that intention, and I intend to implement that change.
On the other hand, I think that here you can make a direct comparison and use the ConfigError exception.
if name.isidentifier() or iskeyword(name):
   raise ConfigError("....")

cainesi: Fixed.

I don't normally use PEP8 names, but will try to remember that for this project.

PR #729, opened 2020-07-28
Added sample code to show the use of ConstantsModule / BUILD_CONSTANTS

cainesi:
I think the title explains everything.

### DIFF
--- a/cx_Freeze/samples/build_constants/hello.py
+++ b/cx_Freeze/samples/build_constants/hello.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from datetime import datetime
+import sys
+from sys import stdout
+
+stdout.write('Hello from cx_Freeze\n')
+stdout.write('The current date is %s\n\n' %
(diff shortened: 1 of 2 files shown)


cx_Freeze/samples/build_constants/hello.py
```diff
@@ -0,0 +1,25 @@
...
+
+stdout.write('Hello from cx_Freeze\n')
+stdout.write('The current date is %s\n\n' %
+             datetime.today().strftime('%B %d, %Y %H:%M:%S'))
+
+stdout.write('Executable: %r\n' % sys.executable)
+stdout.write('Prefix: %r\n' % sys.prefix)
+stdout.write('Default encoding: %r\n' % sys.getdefaultencoding())
+stdout.write('File system encoding: %r\n\n' % sys.getfilesystemencoding())
+
```
marcelotduarte: Why use 'stdout.write' here and print in another place? I think you can use print everywhere.

cainesi: It just came from the code in samples/simple that I copied.  Removed most of it and replaced the remaining ones with prints.

PR #735, opened 2020-07-31
Added a sample to illustrate problem with importlib.util.find_spec

cainesi:


klensy:
https://docs.python.org/3.8/library/importlib.html#importlib.util.find_spec
>if name is for a submodule (contains a dot), the parent module is automatically imported.

Maybe that?

cainesi:
> https://docs.python.org/3.8/library/importlib.html#importlib.util.find_spec
> 
> >if name is for a submodule (contains a dot), the parent module is automatically imported.
> 
> 
> 
> Maybe that?

Yes, indeed, that is exactly the problem. I've been thinking about how to fix it.

marcelotduarte:
I already have a fix. In fact, I rewrote FindModule/LoadModule to use importlib.machinery and remove the deprecated 'imp', and with that solves this and other issues.

cainesi:
> I already have a fix. In fact, I rewrote FindModule/LoadModule to use importlib.machinery and remove the deprecated 'imp', and with that solves this and other issues.

Great! It looks like master is still using imp, so those changes are not merged in yet?

marcelotduarte:
I am finalizing some tests and I will put in a PR for tests for you.

PR #754, opened 2020-08-26
Update README.md

iMonZ:
You get an Error message with just the Python command

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ instead.
 Install by issuing the command
 
 ```
-python -m pip install cx_Freeze --upgrade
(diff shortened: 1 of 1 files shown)


tamenol:
> 
> 
> I get an Error message with just the Python command

FIFY. On most systems nowadays `python` points to `python3` since `python2` is EOL.

marcelotduarte:
@iMonZ I agree with @tamenol. And I add: With virtualenv the current command is also correct, but without a virtual environment, on some systems what you propose is correct.

marcelotduarte:
I updated the docs with more information #796
Thanks!

PR #760, opened 2020-10-04
Added summary_data option to bdist_msi

cainesi:
Option allows author, comments, and keywords data to be added to the MSI file's summary information stream.


marcelotduarte:
cx_Freeze 6.3 has just been released.

PR #775, opened 2020-10-31
Sample for bdist_msi, summary_data option

cainesi:
Adds a sample setup script for using the summary_data option for bdist_msi.

### DIFF
--- a/cx_Freeze/samples/msi_summary_data/hello.py
+++ b/cx_Freeze/samples/msi_summary_data/hello.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from datetime import datetime
+import sys
+from sys import stdout
+
+stdout.write('Hello from cx_Freeze\n')
+stdout.write('The current date is %s\n\n' %
+             datetime.today().strftime('%B %d, %Y %H:%M:%S'))
+
+stdout.write('Executable: %r\n' % sys.executable)
+stdout.write('Prefix: %r\n' % sys.prefix)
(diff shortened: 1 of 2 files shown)


cx_Freeze/samples/msi_summary_data/hello.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from datetime import datetime
+import sys
+from sys import stdout
+
+stdout.write('Hello from cx_Freeze\n')
```
marcelotduarte: Hi.
Why use stdout?
If you have no specific reason, please use print.

marcelotduarte (changes requested):

cx_Freeze/samples/msi_summary_data/hello.py
cainesi: Hi Marcelo--happy to change it.  The only reason I used stdout was because stdout was used in the "simple" sample, and I started from that code when making my sample.

marcelotduarte: There are some too old samples from the epoch of py2 that need to be upgraded....

cainesi: I've fixed it for the msi_summary_data sample.  Happy to go through and fix some of the other samples at some point, if that would be welcome.

marcelotduarte: If you can, thank you!

PR #776, opened 2020-11-01
small refactor

klensy:
Hello.

Few simple changes:
1. in `_LoadModule` don't reopen file if it already opened (if it opened, encoding is known), little speedup
2. in `_FreezeExecutable` `dependent_files` to set type, removing unneeded duplicating
3. in `_GetDependentFiles` use default value, so no need to set empty list few times later
4. in `Freeze` also changed `filesCopied` to set type, as i can't see any usage of it as dict

Few annotations added to code to simplify understanding code and debugging.


klensy:
Ah, type annotations unsupported under python 3.5, so me need to remove it.

marcelotduarte:
Hi, you can maintain the type annotation.

klensy:
This one line should be enough.

klensy:
In few places i see that `dict`s used as `set`s, without using any dict properties: only adding element to dict with `mydict[key]=None` and checking key presence in dict; it seems like some old refactored code.

klensy:
Review please? Thanks.

marcelotduarte:
Thanks!

PR #781, opened 2020-11-11
Cause MSI file to be released at the end of bdist_msi command.

cainesi:
When the bdist_msi command is run from within a larger script, then the .db member was not being released at the end of the command.  As a result the .msi remained locked blocking changes by other programs. This change corrects this.

(In my particular case, I have a script that runs the bdist_msi command (via distutils.core.run_setup) and then uses signtool.exe to sign the resulting msi.  Signtools.exe was failing because the file was still open.)


marcelotduarte:
Thanks!

PR #804, opened 2020-11-25
Update version in __init__.py

ferdnyc:
I noticed that both 6.3 and 6.4 release packages report their version as 6.2, still — the `__version__` in `init.py` isn't being updated. Seems like either a release procedure or script is missing a step to bump that.

Perhaps I should be bumping it to 6.5 here, in fact. Regardless, the point of this PR isn't really to bump the version. It's primarily to call attention to the fact that it's not getting incremented for each release.


marcelotduarte:
I fix the issue with PR #806 and release the new version  6.4.1
Thanks!

PR #811, opened 2020-11-28
refactor 'module finder' to use importlib.machinery

marcelotduarte:
and get rid of deprecated imp

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -3,18 +3,13 @@ Base class for finding modules.
 """
 
 import dis
-import imp
-import io
 import importlib.machinery
-import importlib.util
 import logging
 import os
 import sys
 import opcode
-import pkgutil
-import tokenize
-from typing import Dict, Optional
-import zipfile
+from typing import Dict, List, Optional, Union
+from importlib.abc import ExecutionLoader
 
 from cx_Freeze.common import rebuild_code_object
 import cx_Freeze.hooks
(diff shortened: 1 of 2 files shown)


cx_Freeze/finder.py
```diff
@@ -412,77 +291,98 @@ def _InternalImportModule(self, name, deferredImports):
...
-        if type == imp.PKG_DIRECTORY:
-            return self._LoadPackage(name, path, parent, deferredImports)
-        module = self._AddModule(name, file_name=path, parent=parent)
+    def _load_module(
+        self,
+        name: str,
+        path: Union[str, List[str]],
+        deferredImports: List[str],
+        parent: Optional["Module"] = None,
+    ) -> "Module":
```
klensy: Shouldn't it be `Optional["Module"]`, because function can return `None` in few cases?

marcelotduarte: Optional should be used for arguments.
https://docs.python.org/3/library/typing.html?highlight=typing#typing.Optional

klensy: No, you can use Optional for return type (and other types). For example you can check this if you run mypy for this file (or use pycharm, it will give you warning about return type). https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

klensy: Actually, there places, where `Module` expected, but `Optional['Module']` returned, so here can be bugs.

cainesi: For what it's worth, I think the "Optional" might also be appropriate on the return type.  No affect on the code, but it will help the IDEs.

klensy: @cainesi we talk exactly about return type
```python
) -> "Module":
```

marcelotduarte: The question is: We can or we shall?
I understood that it was not necessary, from the documentation, that talks about an argument. I had also seen a response in the SO that made me continue my understanding.
As I am not a native English speaker, I may have misunderstood. I had understood arguments as parameters.
But I do not see the possibility of "bugs" because it is just a type annotation, and at runtime, it is ignored.
For us, the usefulness is that @cainesi well said.

cainesi: I took a look at the documentation for "Optional".

The reason it talks about arguments is to explain that "Optional" is *not* the same thing as a function argument with a default value. It looks like "Optional" can be used anywhere you are providing a type hint. 

If you have any english questions let me know. I'm up in Canada.

marcelotduarte: > If you have any english questions let me know. I'm up in Canada.

@cainesi, In general, my English is good, and I use Grammarly to give me suggestions. Maybe for writing, my English is better than my Portuguese ;-)
Sometimes I can misunderstand some words that are not well explained, as I previously commented.
Thank you and @klensy for pointing me the mistake.
I hope this is the only problem with this code change.

PR #819, opened 2020-12-04
pin importlib_metadata version

klensy:
pin importlib_metadata to >=3.1.1, as there was error in https://github.com/python/importlib_metadata/issues/261

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,7 @@ keywords =
 [options]
 python_requires = >=3.6
 install_requires =
-    importlib_metadata >= 1.6.1
+    importlib_metadata >= 3.1.1
 # https://setuptools.readthedocs.io/en/latest/setuptools.html#setting-the-zip-safe-flag
 zip_safe = False
 


marcelotduarte:
In py38 we need importlib_metadata too. In py 3.8.3, for example, the bundled lib is buggy. And previous versions do not support files().
I pinned to 1.6.1 because it is the same version as py39, but I forgot to limit it to <=py38.
But with this new bug in importlib_metadata, I think we should investigate what versions after 1.6.1 should be excluded.

klensy:
Ok, i removed python version restriction.

klensy:
But wait,
```python
try:
    import importlib.metadata as importlib_metadata
except ImportError:
    import importlib_metadata
```
at first, bundled version of importlib_metadata imported first, so for python >= 3.8 bundled version never used?

marcelotduarte:
importlib_metadata contains a mechanism that patches it.

klensy:
So no difference ?
```python
try:
    import importlib_metadata
except ImportError:
    import importlib.metadata as importlib_metadata
```

marcelotduarte:
No. From the docs[1]:
"Users of Python 3.8 and beyond are encouraged to use the standard library module when possible and fall back to importlib_metadata when necessary. When imported on Python 3.8 and later, importlib_metadata replaces the DistributionFinder behavior from the stdlib, but leaves the API in tact. "

[1] https://importlib-metadata.readthedocs.io/en/latest/

klensy:
> 
> 
> importlib_metadata contains a mechanism that patches it.

I don't understand. 

```python
try:
    import importlib.metadata as importlib_metadata
except ImportError:
    import importlib_metadata
```
If `try` block runs without exception, then `except` block never executes,so `importlib_metadata` never imported and no patching? It will work, if bundled `importlib.metadata` trying to import `importlib_metadata` inside itself.

marcelotduarte:
This code must also be modified. When I implemented the use of metadata, that's fine. Now, this is inefficient. And as this bug is reported, I see that we may have a bug in the cx-freeze too, because I used a similar technique.

marcelotduarte:
I edited the previous quote, I "ate" a part of it .. ;-)
"Users of Python 3.8 and beyond are encouraged to use the standard library module when possible and fall back to importlib_metadata when necessary. When imported on Python 3.8 and later, importlib_metadata replaces the DistributionFinder behavior from the stdlib, but leaves the API in tact. "

marcelotduarte:
Thanks!

PR #820, opened 2020-12-05
always use importlib_metadata

marcelotduarte:
this is a complement to PR #819

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -16,13 +16,10 @@ import tokenize
 from typing import Dict, Optional
 import zipfile
 
+import importlib_metadata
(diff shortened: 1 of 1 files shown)


klensy:
Btw, this is not the only place where `importlib_metadata` imported, so, it won't became inconsistent?

marcelotduarte:
Where? In the docs? That is for readthedocs.

klensy:
https://github.com/marcelotduarte/cx_Freeze/blob/31dbf4e8b749bbe640922aada1b12b3c7a174923/cx_Freeze/__init__.py#L40

marcelotduarte:
PR #822 
Thanks!

PR #826, opened 2020-12-07
Added windows dependencies resolver [WIP]

klensy:
For windows binaries added simple dependencies resolver
Don't merge yet, as need more testing.

Pros:
* pyside2 (and other hard deps, hopefully) example works without any manual work

Cons:
* bigger size, as some dependencies copied more than 1 time

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -22,6 +22,7 @@ import zipfile
 from .common import ConfigError, get_resource_file_path, process_path_specs
 from .darwintools import DarwinFile, MachOReference, DarwinFileTracker
 from .finder import ModuleFinder
+from .petools import get_pe_dependencies
 
 if sys.platform == "win32":
     import cx_Freeze.util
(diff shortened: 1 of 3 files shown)


marcelotduarte:
I have an option for this based on a code developed in C. But I need to reorganize some code in freeze and do better tests too.
I already know the pefile lib but want a faster option. 
Later, when I have time, I'll put my version in a branch.
I think that we can test these option enabling options like --experimental-pefile and --experimental-winldd

marcelotduarte:
Hi @klensy 
An option developed in C is in the branch `winldd_as_pyd`, that I made as an extension. It is based on https://github.com/greenjava/win-ldd and https://github.com/LRN/ntldd.
But it has a bug (segfault) hard to find. It occurs in very few DLLs (I noted in two), but one is very commonly used: python3.dll. It segfault occurs with the original code of win-ldd too.

Then I take another approach. Use lief, a library made in C++ with bindings for python.

marcelotduarte:
[cx_Freeze 6.10](https://github.com/marcelotduarte/cx_Freeze/releases/tag/6.10) has just been released.

PR #827, opened 2020-12-08
Implemented a "plist_items" option on bdist_mac command

cainesi:
The new option allows the setup script to specify specific extra keys that should be added to the app bundle's Info.plist file (without needing to completely replace the file with a custom plist using the custom_info_plist).

(Incidentally, this was initially inspired by a desire to set NSHighResolutionCapable to True on an app.  Not sure if we should just be setting that key by default at this point?)


cx_Freeze/macdist.py
```diff
@@ -165,6 +171,7 @@ def initialize_options(self):
         self.iconfile = None
         self.qt_menu_nib = False
         self.bundle_name = self.distribution.get_fullname()
+        self.plist_items = None
```
marcelotduarte: This element is a list, see https://github.com/marcelotduarte/cx_Freeze/blob/master/cx_Freeze/dist.py#L186

cx_Freeze/macdist.py
```diff
@@ -177,6 +184,15 @@ def initialize_options(self):
 
     def finalize_options(self):
         self.include_frameworks = normalize_to_list(self.include_frameworks)
+        if self.plist_items is None: self.plist_items = []
```
marcelotduarte: With change in initialize_option this should be changed too, like in https://github.com/marcelotduarte/cx_Freeze/blob/master/cx_Freeze/dist.py#L221

marcelotduarte (changes requested):

cx_Freeze/macdist.py
cainesi: Made the change.  

One thing I'll note: if we change the list option handling to be like in build_exe, then IDEs won't be able to understand that the option members exist on the objection.  For that reason, I would slightly prefer just listing out all the options explicitly in initialize_options()--but I do not care strongly.

marcelotduarte: For IDE you can use typing hints, as PEP 526, without an initial value.

cainesi: Would you be okay listing all the List-options in initialize_options(), to hint their types?

marcelotduarte: I meant, without changing what you've already done,
```
class bdist_mac(Command):
    plist_items: List[...]
```

cainesi: Makes sense.  Done.

cx_Freeze/macdist.py
```diff
@@ -92,6 +93,8 @@ def run(self):
 class bdist_mac(Command):
     description = "create a Mac application bundle"
 
+    plist_items: List[Tuple[str,str]]
+
```
marcelotduarte, line 99: Wouldn't it be nice to add the other elements that you modified? I mean,
include_frameworks: List[str]
include_resources: List[str]

cx_Freeze/macdist.py
```diff
@@ -1,7 +1,8 @@
 from distutils.core import Command
-import os
-import plistlib
-import subprocess
+import os, plistlib, subprocess
```
marcelotduarte: Please do not touch this format, because when I run black I will have to reformat it.

cainesi:
I think all the requested changes have been made, but please let me know if I missed anything.

cx_Freeze/macdist.py
marcelotduarte: Here is pending

cx_Freeze/macdist.py
marcelotduarte: Here is pending

marcelotduarte:
@cainesi This PR has 2 or 3 pending requests and a conflict with your previous PR #834

cainesi:
> @cainesi This PR has 2 or 3 pending requests and a conflict with your previous PR #834

The conflicts are resolved, and I think I addressed the comments above.  So I think this should be good now.

Happy holidays.

marcelotduarte (approved):

marcelotduarte:
Thanks!

cx_Freeze/macdist.py
cainesi: I am happy to make this change.  But I see that the PR is already merged, so not sure if you still want it?

marcelotduarte: I did it in PR #859

PR #828, opened 2020-12-08
Removed rpath-lib-folder option from bdist_mac

cainesi:
This pull would remove the rpath-lib-folder option, which no longer does anything (the option is literally never used in the code) after the recent changes that added the darwintools module.  Previously, that option would determine the value of rpath used when determining dependencies on OSX.  Now, rpath is determined by looking at the values specified in the linking binaries.

The other alternative I can see are:
1. Mark the option as depreciated, and print a warning when a setup script is run that uses the option.  That way old setup scripts will continue to run, but the user will at least be warned that the rpath-lib-folder is no longer doing what it used to do.

2. Allow the option, if set, to override the determination of the rpath in the darwintools module. I suppose in theory someone might have a script that actually depended on the older behaviour, and that was not fixed by the darwintools changes.

### DIFF
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -2,6 +2,7 @@ from distutils.core import Command
 import os
 import plistlib
 import subprocess
+import warnings
 
 from cx_Freeze.common import normalize_to_list
 
@@ -157,7 +158,7 @@ class bdist_mac(Command):
         (
             "rpath-lib-folder",
             None,
-            "replace @rpath with given folder for any files",
+            "[DEPRECIATED]",
         ),
     ]
 
@@ -177,6 +178,11 @@ class bdist_mac(Command):
 
     def finalize_options(self):
         self.include_frameworks = normalize_to_list(self.include_frameworks)
+        if self.rpath_lib_folder is not None:
+            warnings.warn(
+                "rpath-lib-folder is obsolete and will be removed in the next version"
+            )
+
 
     def create_plist(self):
         """Create the Contents/Info.plist file"""
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -158,7 +158,7 @@ class bdist_mac(Command):
         (
             "rpath-lib-folder",
             None,
-            "[DEPRECIATED]",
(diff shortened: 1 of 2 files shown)


marcelotduarte:
> (the option is literally never used in the code)

If this is true, I think that the best is option 1 - Mark the option as depreciated,
You can do like I do in PR #800
but inform that "is obsolete and will be removed in the next major version"
or better text to information that will be removed in 7.0

cainesi:
Done, changed it to mark the option as depreciated.

cx_Freeze/macdist.py
```diff
@@ -157,7 +158,7 @@ class bdist_mac(Command):
         (
             "rpath-lib-folder",
             None,
-            "replace @rpath with given folder for any files",
+            "[DEPRECIATED]",
```
marcelotduarte: I usually see in similar situations, something like: OBSOLETE, DISCONTINUED or DEPRECATED.

cainesi: Sorry, typo.  I deal with "depreciation" a lot in my job day-to-day, so wrote that without thinking.

marcelotduarte: OK. When you have finished, please, remove the draft.

cainesi: Sorry--missed that comment. Would be happy to remove the draft.

(Did you intend to close this pull request unmerged?)

marcelotduarte: I didn't close it, but I can see that you changed the base branch to develop, and when updating my develop, it is recreated, so it closed what was pending ....

cainesi: Ah, got it.  Do you prefer the pull requests on master?

marcelotduarte: It is not a matter of my preference. The develop branch is unstable.

marcelotduarte:
Thanks!
@cainesi When the other 2 patches are complete (I asked for some changes) I will merge them and release a new version.

marcelotduarte:
@cainesi I need you to make a new PR based on the master.

cainesi:
Done.  See #834.

> @cainesi I need you to make a new PR based on the master.

PR #830, opened 2020-12-12
Suppress additional output if --silent has been set.

cainesi:
This makes the script a bit closer to silent when "silent" operation has been specified.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -56,7 +56,6 @@ class Freezer:
         includeFiles: Optional[List] = None,
         zipIncludes: Optional[List] = None,
         silent: bool = False,
-        noMissingWarnings: bool = False,
         metadata: Optional[DistributionMetadata] = None,
         includeMSVCR: bool = False,
         zipIncludePackages: Optional[List[str]] = None,
@@ -86,7 +85,6 @@ class Freezer:
         self.includeFiles = process_path_specs(includeFiles)
         self.zipIncludes = process_path_specs(zipIncludes)
         self.silent = silent
-        self.noMissingWarnings = noMissingWarnings
         self.metadata = metadata
         self.zipIncludePackages = list(zipIncludePackages or [])
         self.zipExcludePackages = list(zipExcludePackages or [])
@@ -594,8 +592,7 @@ class Freezer:
 
         if not self.silent:
             self._PrintReport(fileName, modules)
-        if not self.noMissingWarnings:
-            finder.ReportMissingModules()
+        finder.ReportMissingModules()
 
         targetDir = os.path.dirname(fileName)
         self._CreateDirectory(targetDir)
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -56,6 +56,7 @@ class Freezer:
         includeFiles: Optional[List] = None,
         zipIncludes: Optional[List] = None,
         silent: bool = False,
+        noMissingWarnings: bool = False,
         metadata: Optional[DistributionMetadata] = None,
         includeMSVCR: bool = False,
         zipIncludePackages: Optional[List[str]] = None,
@@ -85,6 +86,7 @@ class Freezer:
         self.includeFiles = process_path_specs(includeFiles)
         self.zipIncludes = process_path_specs(zipIncludes)
         self.silent = silent
+        self.noMissingWarnings = noMissingWarnings
         self.metadata = metadata
         self.zipIncludePackages = list(zipIncludePackages or [])
         self.zipExcludePackages = list(zipExcludePackages or [])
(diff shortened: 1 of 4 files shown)


cx_Freeze/freezer.py
```diff
@@ -572,7 +572,9 @@ def _WriteModules(self, fileName, finder):
 
         if not self.silent:
             self._PrintReport(fileName, modules)
-        finder.ReportMissingModules()
+
+        if not self.silent:
```
marcelotduarte: An indentation is sufficient to place it in the silent block. But, I wonder why the original author did not do this.

marcelotduarte (changes requested):

cx_Freeze/freezer.py
cainesi: I know. I put it in a separate block as a hint to people in the future that they may want to consider different conditions for calling the ReportMissingModules() method.

I guess that the original author probably did not make the method conditional on "no self.self", because he viewed the missing modules as an error that the user should be informed about in any case. However, the problem is that ReportMissingModules() reports many "missing" modules no matter what-- for example, platform specific modules that are not available on other modules.

We could put in a safe list, of modules that we know are okay to be missing, and ignore those ones.  But that seems a lot more complicted.

marcelotduarte: > reports many "missing" modules no matter what

It also bothers me and I've tried to reduce it, because sometimes it helps to identify what is wrong, but with a very long list, it doesn't help at all ...

cainesi: Yeah. 

I think if someone knows enough to be helped by the list of missing modules, then they would also know enough to turn off "silent" when they encounter a problem.

cx_Freeze/macdist.py
```diff
@@ -31,9 +31,13 @@ class bdist_dmg(Command):
     def initialize_options(self):
         self.volume_label = self.distribution.get_fullname()
         self.applications_shortcut = False
+        self.silent = None
```
marcelotduarte, line 38: Is this option documented for this class?

cx_Freeze/freezer.py
cainesi: Incidentally, I was looking at distutils.  The way it handles the verbosity is by setting a global reporting threshold in the distutils.log module, and directing all the output through distutils.log.log() (and some sibling functions).  Should cx_Freeze standardize on doing that as well?

cx_Freeze/macdist.py
```diff
@@ -31,9 +31,13 @@ class bdist_dmg(Command):
...
 
     def finalize_options(self):
-        pass
+        self.set_undefined_options(
+            "build_exe",
```
marcelotduarte: Why did you add the option build_exe? Can you also add documentation?

cx_Freeze/freezer.py
cainesi: Are you okay having ReportMissingModules() silenced?  If you still want ReportMissingModules() to run by default, we could add another option specifically for it ("--hide-missing"?).

marcelotduarte (changes requested):
I was a little reluctant about this change. I think you can leave it as it was and in another patch implement this new option.

cx_Freeze/macdist.py
cainesi: My thinking was that bdist_dmg would just pick up the behavior specified for build_exe (which already has a silent option), without really defining a separate option for bdist_dmg.  But if you'd prefer I explicitly define and document a separate option for bdist_dmg, I am happy to do that instead.

marcelotduarte: Please, document it.

cainesi: > I was a little reluctant about this change. I think you can leave it as it was and in another patch implement this new option.

Done.  There is now a no_missing_warnings option on build_exe to suppress the list of missing modules.

One thought, for the future--I think if we replace the print & shutil.copyfile/copytree code with calls to distutils.dir_util.copy_tree and distutils.file_util.copy_file, we could piggy back off the distutils logging code and the distutils verbosity level would also affect cx_freeze's output.  Too busy to do that at the moment.

cx_Freeze/macdist.py
cainesi: Fixed.

cainesi:
> I was a little reluctant about this change. I think you can leave it as it was and in another patch implement this new option.

Sorry-- I misunderstood this request at first.  I've now separated out the "no_missing_warnings" into a separate pull request (#883).  This PR now only adds a silent flag for bdist_dmg and suppresses some file copying messages in build_exe.

marcelotduarte:
@cainesi You removed the docs in doc/src/distutils.rst, can you add it?
image

cainesi:
> @cainesi You removed the docs in doc/src/distutils.rst, can you add it?

Sorry that was a mistake.  I had only meant to remove the "no_missing_warnings" options, which is moved to a different pull request.  I have put the documentation of the bdist_dmg "silent" option back in.

marcelotduarte:
Thanks!

PR #858, opened 2020-12-30
The name of the target executable can be modified after the build

### LINKED ISSUE #703: Allow changes in executable filename
The bundled application currently looks at the name of the executable to determine which python module to load to start the program (this is from __startup__.py):

    def run():
        baseName = os.path.normcase(os.path.basename(sys.executable))
        name, ext = os.path.splitext(baseName)
        module = __import__(name + "__init__")
        module.run()

This means that the program breaks if the executable filename changes, which is a bit unexpected.  Would people be open to a change to instead store the name of the relevant module in the BUILD_CONSTANTS module?

marcelotduarte:
(only if one executable is built)

### DIFF
--- a/cx_Freeze/initscripts/__startup__.py
+++ b/cx_Freeze/initscripts/__startup__.py
@@ -44,13 +44,18 @@ def run():
     try:
         module = __import__(name + "__init__")
     except ModuleNotFoundError:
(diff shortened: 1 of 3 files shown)


cainesi:
Thanks!  I took a look at the code and it makes sense to me.  It's more elegant than what I had planned.  I'll test it out when I get some free time.

PR #883, opened 2021-01-14
Implemented multi levels for build_exe silent option

cainesi:
If set, the new option will suppress printing of the list of missing modules.

### DIFF
--- a/cx_Freeze/dist.py
+++ b/cx_Freeze/dist.py
@@ -144,7 +144,15 @@ class build_exe(distutils.core.Command):
             "and place in the file system instead (or * for all) "
             "[default: *]",
         ),
-        ("silent", "s", "suppress all output except warnings"),
+        (
+            "silent=",
+            "s",
+            "suppress output from build_exe command.  "
+            "level 0: get all messages; [default]"
+            "level 1: suppress information messages, but still get warnings; [default if flag specified with no specific level]"
+            "level 2: suppress missing missing-module warnings "
+            "level 3: suppress all warning messages"
+        ),
     ]
     boolean_options = ["no-compress", "include_msvcr", "silent"]
 
@@ -220,8 +228,14 @@ class build_exe(distutils.core.Command):
         self.set_undefined_options("build", ("build_exe", "build_exe"))
         self.optimize = int(self.optimize)
 
-        if self.silent is None:
-            self.silent = False
+        if self.silent is None: self.silent = 0
+        elif self.silent is False: self.silent = 0
+        elif self.silent is True: self.silent = 1
+        elif isinstance(self.silent, int): pass
+        elif isinstance(self.silent, str):
+            try: self.silent = int(self.silent)
+            except ValueError: self.silent = 1
+        else: self.silent = 1
 
         # Make sure all options of multiple values are lists
         for option in self.list_options:
@@ -252,7 +266,7 @@ class build_exe(distutils.core.Command):
             binIncludes=self.bin_includes,
             binExcludes=self.bin_excludes,
             zipIncludes=self.zip_includes,
-            silent=self.silent,
+            silentLevel=self.silent,
             binPathIncludes=self.bin_path_includes,
             binPathExcludes=self.bin_path_excludes,
             metadata=metadata,
--- a/cx_Freeze/dist.py
+++ b/cx_Freeze/dist.py
@@ -272,7 +272,7 @@ class build_exe(distutils.core.Command):
             binIncludes=self.bin_includes,
             binExcludes=self.bin_excludes,
             zipIncludes=self.zip_includes,
-            silentLevel=self.silent_setting,
+            silent=self.silent_setting,
             binPathIncludes=self.bin_path_includes,
             binPathExcludes=self.bin_path_excludes,
             metadata=metadata,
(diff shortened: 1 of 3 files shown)


cx_Freeze/dist.py
```diff
@@ -140,6 +140,8 @@ class build_exe(distutils.core.Command):
             "[default: *]",
         ),
         ("silent", "s", "suppress all output except warnings"),
+        ("no-missing-warnings", "n", "suppress warnings concerning missing modules"),
```
marcelotduarte: Sorry for the delay. I was slow to have a different idea from yours because I didn't want a 'no' option.
I was wondering if we could continue using the 'silent' option with extended use. Today it is a boolean option and maybe it could become a multiple options:
--silent (-s)
--silent=info (or true)
--silent=warnings (missing modules)
--silent=all (info+warnings)

marcelotduarte (changes requested):

cx_Freeze/dist.py
cainesi: Sounds fine to me.  I can revise the PR. 

Just to confirm I do this right, you are thinking that the levels would be
- 0 (-s flag not set): get all the messages
- 1: get all warning messages, but none of the "info" messages
- 2: get warning messages, _except_ the missing-module warnings
- 3: suppress all warnings (would still get failure messages)

Hope you're enjoying the long weekend.

marcelotduarte: Sounds fine.

Unfortunately, in my country, things are not good, so we just have to stay tuned.

cainesi: Okay, I have implemented the new "silent" option as suggested.

marcelotduarte (approved):

cx_Freeze/dist.py
marcelotduarte: It looks like it looks really good.

marcelotduarte:
runing simple sample
--silent=0 and --silent=1 ok
--silent
```
python setup.py build_exe --silent
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --silent requires argument
```
--silent=2
error: invalid truth value '2'
--silent=3
error: invalid truth value '3'

cainesi:
Sorry, I never use the command line, so hadn't tested that!

It looks like there is a little problem, and a bigger problem.

Little problem: we should take the option out of boolean_options, so that it can be given values other than true/false on the command line.

Bigger problem: it looks like an option either (i) has no argument, or (ii) requires an argument.  As far as I can see, you cannot have an option that may or may not have an argument (though the documentation of distutils is not great!).  If we force silent to have an argument, we would break backward compatibility.

So I think what we should do is split this into two (overlapping) options: silent-level (with an argument, like what is in the PR now), and silent (with no argument, but equivalent to silent-level=1).

I've revised PR accordingly.

cx_Freeze/freezer.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         binPathExcludes: Optional[List] = None,
         includeFiles: Optional[List] = None,
         zipIncludes: Optional[List] = None,
-        silent: bool = False,
+        silentLevel: int = 0,
```
marcelotduarte: You can maintain the variable name 'silent' in freeze.py, accepting bool or int, because there are users that use the API

doc/src/distutils.rst
```diff
@@ -179,7 +179,17 @@ It can be further customized:
...
+   * - silent_level
+     - suppress output from freeze process; can provide a value to specify
+       what messages should be suppressed, with the possible values being:
+       0 - do not suppress any output [default];
+       1 - suppress information messages;
+       2 - also suppress missing-module warning messages;
+       3 - also suppress all other warning messages.
+
+.. versionchanged:: 6.7
+    Added multiple levels for ``silent`` option.
```
marcelotduarte: Added ``silent_level`` option.

cx_Freeze/freezer.py
marcelotduarte: You're so fast. ;-)
but I need to accept bool or int because there are users that use the API

cx_Freeze/dist.py
```diff
@@ -144,7 +144,16 @@ class build_exe(distutils.core.Command):
...
+        ("silent", "s", "suppress all output except warnings (equivalent to --silent-level=1)"),
+        (
+            "silent-level=",
+            None,
+            "suppress output from build_exe command.  "
+            "level 0: get all messages; [default]"
+            "level 1: suppress information messages, but still get warnings; (equivalent to --silent)"
+            "level 2: suppress missing missing-module warnings "
+            "level 3: suppress all warning messages"
+        ),
```
marcelotduarte, line 156: It is a good way out.

cx_Freeze/freezer.py
cainesi: Just happen to be at my computer.  I think I fixed the Freezer parameter now.

marcelotduarte:
ok, I'll do manual tests and CI test in all pythons.

marcelotduarte:
Thanks!

Issue #873, opened 2021-01-12
Mac error when freeze PySide2 app with virtual environment activated 


ktmlleska:
Hello, 
I'm in a python3k virtual environment and I'm to build a package for mac that uses PySide2. When building with bdist_mac with cx_Freeze 6.5 I receive the following error:

`cx_Freeze.darwintools.DarwinException: resolveRPath() failed to resolve path: @rpath/QtCore.framework/Versions/5/QtCore`

Which I understand that it happens because the @rpath is pointing to: `/Users/MyUser/.pyenv/versions/3.7.6/envs/MyVirtualEnv/lib/python3.7/site-packages/PySide2`
when trying to look for the PySide framework. 

Instead it should be looking to: `/Users/MyUser/.pyenv/versions/3.7.6/envs/MyVirtualEnv/lib/python3.7/site-packages/PySide2/Qt/lib` which is where the framework files exist.

Does anyone knows any solution to this or found the same issue?

Thanks!

marcelotduarte:
Hi
Can you test the PR #887 (if you have a C compiler):
pip install -U git+https://github.com/marcelotduarte/cx_Freeze.git@refs/pull/887/head

ktmlleska:
Thanks I'll test it.

ktmlleska:
Hey, thanks! it solves the PySide errors, but still failing with the following error:
`error: [Errno 2] No such file or directory: '/usr/local/opt/gdbm/lib/libgdbm.4.dylib'`
In that folder the only files I have are libgdbm.6.dylib, libgdbm_compat.4.dylib and libgdbm.dylib
But I guess this is an unrelated error, so it seems to me that this issue is solved.
Thank you again.

marcelotduarte:
@cainesi He tested your PR too.

cainesi:
> Hey, thanks! it solves the PySide errors, but still failing with the following error:
> 
> `error: [Errno 2] No such file or directory: '/usr/local/opt/gdbm/lib/libgdbm.4.dylib'`
> 
> In that folder the only files I have are libgdbm.6.dylib, libgdbm_compat.4.dylib and libgdbm.dylib
> 
> But I guess this is an unrelated error, so it seems to me that this issue is solved.
> 
> Thank you again.

Just to confirm – the program runs okay unfrozen, but the freeze fails with the message above￼?

Would you mind uploading all the output from cx_freeze so I can see what is happening. Might still be something I can fix?

Thanks.

[...]

Issue #846, opened 2020-12-26
Crash running frozen App with Qt5 and MacOS Catalina


ghost:
Hi all,

When building and packaging the following basic Qt5 code:

File Test.py
```
from PyQt5.QtWidgets import QApplication, QLabel
app = QApplication([])
label = QLabel('Hello World')
label.show()
app.exec_()

```

using the file setup.py:

```
from cx_Freeze import setup, Executable

setup(
    name = "salut",
    version = "0.1",
    description = "Ce programme vous dit bonjour",
    options = {"build_exe" : {},
               "bdist_dmg": {}},
    executables = [Executable(script="Test.py")],
)

```

and running `python3 setup.py bdist_dmg`

with installed:
cx_freeze 6.4.2
pyQt5 5.15.2
python 3.9.1

I get a segfault when running the Test application located into .app directory. 

Here is the stacktrace:
```
0   com.apple.CoreFoundation      	0x00007fff368593f8 CFBundleCopyBundleURL + 9
1   QtCore                        	0x0000000102791c3c QLibraryInfo::location(QLibraryInfo::LibraryLocation) + 1260
2   QtCore                        	0x0000000102791e8e QLibraryInfo::location(QLibraryInfo::LibraryLocation) + 1854
3   QtCore                        	0x0000000102964716 QCoreApplication::libraryPathsLocked() + 1094
4   QtCore                        	0x0000000102960ebb QCoreApplication::libraryPaths() + 43
5   QtCore                        	0x00000001029504b4 QFactoryLoader::update() + 36
6   QtCore                        	0x0000000102951d70 QFactoryLoader::QFactoryLoader(char const*, QString const&, Qt::CaseSensitivity) + 224
7   QtGui                         	0x00000001021a0022 0x102174000 + 180258
8   QtGui                         	0x00000001021a0351 QPlatformIntegrationFactory::keys(QString const&) + 641
9   QtGui                         	0x00000001021acda9 QGuiApplicationPrivate::createPlatformIntegration() + 1673
10  QtGui                         	0x00000001021ae1fb QGuiApplicationPrivate::createEventDispatcher() + 27
11  QtCore                        	0x000000010296091e QCoreApplicationPrivate::init() + 1246
12  QtGui                         	0x00000001021a8a49 QGuiApplicationPrivate::init() + 57
13  QtWidgets                     	0x0000000101bd0daa QApplicationPrivate::init() + 26
[...]

PR #887, opened 2021-01-17
Improvements to dependency resolution on Darwin

cainesi:
This PR makes several improvements to the freeze process on OSX.
1. The most critical one is to fix a bug in MachOCommand._getMachOCommands() which would cause the last command in a MachO file not to be correctly processed.
2. When a load command based on @rpath cannot be resolved, DarwinFileTracker.finalizeReferences() will now try to guess which file was intended to be referenced (and throw an exception if it cannot).  This was motivated by my having to deal with a package that had some  incorrect rpaths.
3.  Improvements to error reporting.

If you want, I can also go through darwintools.py and make it all snake_case instead of camelCase.  Let me know.

### DIFF
--- a/cx_Freeze/darwintools.py
+++ b/cx_Freeze/darwintools.py
@@ -140,6 +140,39 @@ class DarwinFile:
             l.append("   {} -> {}".format(rp, self.libraryPathResolution[rp]))
         return "\n".join(l)
 
+    def fileReferenceDepth(self) -> int:
+        """Returns how deep this Mach-O file is in the dynamic load order."""
+        if self.referencingFile is not None: return self.referencingFile.fileReferenceDepth() + 1
+        return 0
+
+    def printFileInformation(self):
+        """Prints information about the Mach-O file."""
+        print(f'[{self.fileReferenceDepth()}] File: "{self.originalFilePath}"')
+        print("  Commands:")
+        if len(self.commands) > 0:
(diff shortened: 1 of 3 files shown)


marcelotduarte:
Hi! I'll look at it later. Maybe we can merge at the next release.
> If you want, I can also go through darwintools.py and make it all snake_case instead of camelCase. Let me know.

For new variables or variables that need to change the name, you can change to snake_case in the same patch, for others I prefer a separated patch. Mainly, because this is code that I didn't touch, except for domestic reasons, and can be confused by many changes. ;-)

There are some issues about @rpath (#873, #831, ...), this PR can resolve this?

cainesi:
I expect this PR would resolve those problems in #873 and #831 (and maybe also #846?), though I cannot be sure without looking at the specific files involved.  But I was also having problems with PySide (similar to #873) which is what led to the PR.  I think it would make sense to ask those people to test the PR to make sure if works for then.

I'll make the name changes in a separate PR once this is resolved and I have some time.

marcelotduarte:
Thanks!
@cainesi, I intend to put some samples to be tested as soon as a PR is sent, however, I still have some problems with the macOS part. It is not caused by this PR, since it was already occurring, but I think that only you can solve them.
Later, I'll open an issue.

cainesi:
Sure, I'll do my best! Just let me know what I should look at one it's available.

PR #915, opened 2021-02-13
Temporary fix for darwintools.py error.

cainesi:
As a temporary hack, this change will simply print an error (rather than failing) if the freeze attempts to copy two dynamic libraries to the same location.  This will effectively result in the first copy of the library being used.


marcelotduarte:
Thanks!

PR #922, opened 2021-02-18
Remove subfolders belonging to excluded modules

DraXus:
Currently all subfolders belonging to a module are copied using `shutil.copytree`. This is an issue if any of the submodules are excluded.

The proposed solution in this pull request is to remove the right subfolders using `shutil.rmtree`.

Also fixes a bug in `Freeze` where `self.excludeModules` was never initialized.


cx_Freeze/freezer.py
```diff
@@ -749,7 +757,7 @@ def _WriteModules(self, fileName, finder):
 
     def Freeze(self):
         self.finder = None
-        self.excludeModules = {}
+        self.excludeModules = self.excludes
```
marcelotduarte: Sorry for the delay. I think you found an issue that I already suspected, but not checked it yet!
But, here, if you initialize excludeModules with excludes, only the user excludes will be excluded. There are excludes from hooks (See hooks.py lines with ExcludeModule).
I think that the correct is use self.finder.excludes, so all excludes will be excluded.

cx_Freeze/freezer.py
```diff
@@ -648,6 +648,14 @@ def _WriteModules(self, fileName, finder):
...
                     )
 
+                    # remove the subfolders which belong to excluded modules
+                    excludedFolders = [m[len(module.name) + 1:].replace(".", os.sep)
+                                       for m in self.excludeModules if m.startswith(module.name)]
+                    for folder in excludedFolders:
+                        folderToRemove = os.path.join(targetPackageDir, folder)
+                        print("Removing", folderToRemove + "...")
+                        shutil.rmtree(folderToRemove)
+
```
marcelotduarte, line 682: Can you use the ignore patterns option of copytree? So you eliminate the copy and delete.
For this case, you should implement a function to extend what actually we are. But this is not a priority. Can be made in another PR if you do not have time.

marcelotduarte (changes requested):

cx_Freeze/freezer.py
DraXus: You're very much right, thanks! I've changed it to `self.finder.excludes`

marcelotduarte: There is a conflict. Can you resolve it? And, I think you can use it directly in the WriteModules and delete this self.excludeModules. It is used only for this.

cx_Freeze/freezer.py
DraXus: I've actually tried using ignore patterns from copytree, but it won't work as expected. I'll explain why.

The `ignore` parameter from `copytree` expects to receive a callable with two arguments:
* The directory being visited
* The list of contents returned by `os.listdir()` (i.e. bare file and folder names)

Let's have this example:
- Main package name: `my_package`
- Folder structure: 
  - my_package
    - folder_a
       - subfolder_1
       - subfolder_2
       - subfolder_3
       - additional files
- Package to exclude: `my_package.folder_a.subfolder_1`

By default, `copytree` will copy all content under `my_package`. If we would like to use the ignore function, we would need to exclude the whole `folder_a` (as it will be listed by os.listdir) and there's no way of excluding `folder_a/subfolder_1`.

I've spent quite a few time trying different approaches, but the only way I found to make it properly is by deleting the folders once they're copied. I'm open to more ideas if anyone else has a better way.

cx_Freeze/freezer.py
DraXus: Done :)

cx_Freeze/freezer.py
marcelotduarte: If you join the visited directory with the returned by listdir, then you have a fullname to compare and see that it cannot be copied, so you can wait for the next round, no?

DraXus: No, I don't think so. In that example, we only have the full name when we are visiting `folder_a`. At that point, everything has been copied in the previous visit of `my_package`.

marcelotduarte: Ok. Later I test it and merge. Thanks!

DraXus: Great, thanks! Please let me know if you find any issues. And thanks for supporting this excellent tool :)

marcelotduarte:
@DraXus 
I did some tests and it is giving errors in the build. It is not directly in your patch but as a result of it. I tried to find the source of the problems, but I still haven't succeeded.
If you test with a simple application it passes, but if you use one of the [samples](https://github.com/marcelotduarte/cx_Freeze/tree/main/cx_Freeze/samples), like pandas or matplotlib it will give the error.

Using pandas sample:
error: [Errno 2] No such file or directory: 'build\\exe.win-amd64-3.8\\lib\\numpy\\random\\_examples\\cffi\\extending.pyc'
Error: Process completed with exit code 1.

DraXus:
@marcelotduarte I think I identified the root cause of the issue.

When calling `load_numpy()` hook, it will first load the whole `numpy` and then exclude `numpy.random._examples`. However, it won't exclude its submodules due to the implementation of `finder.ExcludeModule`.
image

We could improve `finder.ExcludeModule` so it will exclude all the submodules when giving a module. What do you think?

DraXus:
@marcelotduarte I just tried this code and it solves the issue with the pandas test. Do you have some automated script to run over all tests?
```
    def ExcludeModule(self, name: str) -> None:
        """Exclude the named module from the resulting frozen executable."""
        re_name = name.replace(".", "\.")
        regex = re.compile(fr"^{re_name}(\..*)?$")
        matches = [regex.fullmatch(t) for t in self._modules]
        modules_to_exclude = [match.string for match in matches if match is not None]
        for m in modules_to_exclude:
            self.excludes[m] = None
            self._modules[m] = None
```

marcelotduarte:
Yes, I have a semi-automated that uses the samples as tests, and tests in 3 SO.

marcelotduarte:
@DraXus It passes in my tests, except that I have one regression using cxfreeze script on windows.
What SO you are using?

DraXus:
@marcelotduarte I use Windows 10. Which is the failing test?

marcelotduarte:
You are on the right path but have some fails. If you run cxfreeze script directly you can see the error. But I found the cause of the fail. 
For example, the cxfreeze script can use a `__main__` module, which can be used because cx_Freeze creates its proper module.
Then it is excluded in the hooks initialize function. But after your path, a long list of excluded modules is gone.
Another test that you can make is changing the position of ExcludeModule in load_numpy to be before the IncludePackage, and the error should happens.

DraXus:
> If you run cxfreeze script directly you can see the error.

I've never used the cxfreeze script, could you please provide the run command you're using so I can replicate on my side?

> Another test that you can make is changing the position of ExcludeModule in load_numpy to be before the IncludePackage, and the error should happens.

I've tried this but it doesn't give any error on the pandas test.

marcelotduarte:
> I've never used the cxfreeze script, could you please provide the run command you're using so I can replicate on my side?

On command line, use the sample simple: cxfreeze hello.py

> I've tried this but it doesn't give any error on the pandas test.

No error really, but it continues to copy the excluded files.

marcelotduarte:
You can use the pandas sample to test: cxfreeze test_pandas.py

DraXus:
Thanks, I've managed to replicate the error. I'll take a look and see if I can fix it 👍

DraXus:
I found the issue. If the module name is not in the list of current modules, it won't add it to the excludes. I've added the module name to make sure it's added.

```
    def ExcludeModule(self, name: str) -> None:
        """Exclude the named module and its submodules from the resulting frozen executable."""
        re_name = name.replace(".", "\.")
        regex = re.compile(fr"^{re_name}(\..*)?$")
        matches = [regex.fullmatch(t) for t in self._modules]
        modules_to_exclude = [match.string for match in matches if match is not None] + [name]
        for m in modules_to_exclude:
            self.excludes[m] = None
            self._modules[m] = None
```

Please run another round of tests and if it's good I'll add it to the PR.

marcelotduarte:
In a manual test, it works! I started the CI to test.

marcelotduarte:
Tests ok. 
I simplified it a bit:
```
    def ExcludeModule(self, name: str) -> None:
        """Exclude the named module from the resulting frozen executable."""
        modules_to_exclude = [name] + [
            mod for mod in self._modules if mod.startswith(f"{name}.")
        ]
        for mod in modules_to_exclude:
            self.excludes[mod] = None
            self._modules[mod] = None
```
I'll run the tests again.

marcelotduarte:
Tests ok. Can you update the PR?

DraXus:
Thanks, much better without regex :) I've updated the PR.

marcelotduarte:
Thanks!

marcelotduarte:
cx_Freeze 6.6 has just been released.

PR #923, opened 2021-02-20
zip_exclude_packages doesn't exclude a simple module

Mystic-Mirage:
This is not a true PR. Just to show a problem I faced.

I want a simple module to be excluded from library.zip.

Expected:

- to_include.pyc is inside of library.zip
- to_exclude.pyc is outside of library.zip

Actual:
both to_include.pyc 

- and to_exclude.pyc is inside of library.zip

### DIFF
--- a/cx_Freeze/samples/zip_manipulations/hello.py
+++ b/cx_Freeze/samples/zip_manipulations/hello.py
@@ -1,0 +1,7 @@
+#!/usr/bin/env python
+
+import to_include
+import to_exclude
+
+
+print(to_include, to_exclude)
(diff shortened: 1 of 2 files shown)


marcelotduarte:
This option (zip_exclude_packages) works as expected. These are the concepts:
1) [Modules](https://docs.python.org/3/tutorial/modules.html?highlight=module#modules)
2) [Packages](https://docs.python.org/3/tutorial/modules.html?highlight=module#packages)

marcelotduarte:
If you move the contents of to_exclude.py to `to_exclude/__init__.py` you can exclude the package.

Mystic-Mirage:
@marcelotduarte Thank you for your answer

PR #926, opened 2021-02-20
fix plistlib.load call in macdist [py39]

marcelotduarte:


bubio:
Thank you!
It worked fine.

Issue #925, opened 2021-02-20
Failed to create DMG file (applications_shortcut=True)


bubio:
applications_shortcut: If set to True, DMG file creation will fail.

`
hdiutil: create failed - No such file or directory
error: creation of the dmg failed
`

Environment:
- Python 3.9.1
- macOS Big Sur 11.2.1
- Japanese !!

bubio:
macdist.py
  class bdist_dmg(Command):
    def buildDMG(self):

```python
scriptargs = [
  "osascript",
  "-e",
  'tell application "Finder" to make alias \
                file to POSIX file "/Applications" at POSIX file "%s"'
  % os.path.realpath(self.buildDir),
]

if os.spawnvp(os.P_WAIT, "osascript", scriptargs) != 0:
  raise OSError("creation of Applications shortcut failed")
```

Use Applescript's "make alias", the localized folder name will be used.
For example, in a Japanese environment, the name will be "アプリケーション".

For this reason
```python
createargs.append("-srcfolder")
createargs.append(os.path.join(self.buildDir, "Applications"))
```
"Applications" folder specified above will not be found and the hdiutil command will fail to execute.

bubio:
The option "-srcfdolder" in the hdiutil command is specified twice, which also looks strange.

A good way to do this is to create a source folder (e.g. "dist") and create links to the App Bundle and Applications folders in it.

In my environment, the following code worked well.
```python
def buildDMG(self):
        # Remove DMG if it already exists
        if os.path.exists(self.dmgName):
            os.unlink(self.dmgName)

        # Make dist folder
        import shutil

        self.dist_dir = os.path.join(self.buildDir, "dist")
        if os.path.exists(self.dist_dir):
            shutil.rmtree(self.dist_dir)
        self.mkpath(self.dist_dir)

        # Copy App Bundle
        dest_dir = os.path.join(self.dist_dir, os.path.basename(self.bundleDir))
        self.copy_tree(self.bundleDir, dest_dir)

        createargs = [
            "hdiutil",
            "create",
        ]
        if self.silent:
            createargs += ["-quiet"]
        createargs += [
            "-fs",
            "HFSX",
            "-format",
            "UDZO",
            self.dmgName,
            "-imagekey",
            "zlib-level=9",
            "-srcfolder",
            self.dist_dir,
            "-volname",
            self.volume_label,
        ]

        if self.applications_shortcut:
[...]

PR #927, opened 2021-02-21
Failed to create DMG file (applications_shortcut=True) #925

bubio:
- Create the "build/dist" folder.
- Put a symlink to the App Bundle and /Applications folder in "build/dist".
- Use os.symlink instead of Applescript for the symlink to the /Appilcations folder.


marcelotduarte:
Thanks!

PR #954, opened 2021-03-04
Add sample for pyzmq

Dereas:
Simple example of a ZMQ server-client setup with request reply pattern where the client sends a 'Ping' string an the Server responds with a 'Pong' string.


marcelotduarte:
Thanks!

Issue #1019, opened 2021-04-13
Realoding msilib breaks passing `msilib.Binary` objects into `bdist_msi`’s `data` parameter


Cimbali:
### Bug description

On windows, the `data` option passed to `bdist_msi` is used to pass in arbitrary data to put in the installer’s tables. To add stream data, the correct way is to wrap its path in a `msilib.Binary`, as is done with the install icon:

https://github.com/marcelotduarte/cx_Freeze/blob/f5be487e5db58d2a482b272313e6809a06b7e3d0/cx_Freeze/windist.py#L555-L559

However, msilib recognizes these objects with `isinstance()`:

https://github.com/python/cpython/blob/eb77133564d74eb09ed63872a69b9827d4841b49/Lib/msilib/__init__.py#L113-L116

Thus after reloading msilib, `msilib.Binary` objects that were created before the reload are not recognized

https://github.com/marcelotduarte/cx_Freeze/blob/f5be487e5db58d2a482b272313e6809a06b7e3d0/cx_Freeze/windist.py#L819

### Reproduce 

Add any `msilib.Binary` data in the `setup()`’s options, e.g.:

```python
setup(data={
   'Icon': [('SecondIcon', msilib.Binary('path\to\icon')]
})
```
This causes a `TypeError: Unsupported type Binary` to be thrown inside add_config’s add_data.

https://github.com/marcelotduarte/cx_Freeze/blob/f5be487e5db58d2a482b272313e6809a06b7e3d0/cx_Freeze/windist.py#L128-L129

The path doesn’t even have to exist for the error to be thrown.

### Expected behavior

The data is added to the `msi` installer.

### Platform Information

- Windows-10-10.0.17763-SP0
- 64 bits
- Python 3.8.9 in msys2 / mingw64
- cx_Freeze 6.5.3

-----

I’m more than happy to provide a PR for this but there’s a number of ways this could be fixed and I’m not sure which is preferable:
- wrap paths in some `cx_Freeze` object and only make them into `msiilb.Binary` when calling `add_data()`
- (just) before calling `add_data()`, iterate over arguments to identify old `msilib.Binary` objects and re-create them as
  ```python3
  msilib.Binary(old_bin.name)
  ```
- something else I didn’t think of?

marcelotduarte:
My preference is the first option.  Do you think to create an option like binary-data?
@cainesi I see that you have knowledge in this area. Can you suggest?

cainesi:
[...]

PR #1024, opened 2021-04-14
Transform filename to msilib.Binary for binary data

Cimbali:
Only do this for columns marked binary, see #1019.


marcelotduarte:
Can you add the use of this new option in the 'icon' sample to facilitate testing?

Cimbali:
I’ve made a new sample of a realistic case where someone could want to use this (i.e. an icon different from the installer icon and from the executable icon), which also showcases how to add a row in another table.

marcelotduarte:
runing bdist_msi I get:
```
creating dist
  File "setup.py", line 28, in <module>
    setup(
  File "c:\testes\cx_freeze\cx_Freeze\dist.py", line 397, in setup
    distutils.core.setup(**attrs)
  File "C:\Python38-64\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Python38-64\lib\distutils\dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "C:\Python38-64\lib\distutils\dist.py", line 985, in run_command
    cmd_obj.run()
  File "c:\testes\cx_freeze\cx_Freeze\windist.py", line 847, in run
    self.add_config(fullname)
  File "c:\testes\cx_freeze\cx_Freeze\windist.py", line 139, in add_config
    data = [
  File "c:\testes\cx_freeze\cx_Freeze\windist.py", line 140, in <listcomp>
    row[:col] + [msilib.Binary(row[col])] + row[col + 1:]
TypeError: can only concatenate tuple (not "list") to tuple
```

Cimbali:
Ah my bad. The ideal would be `(*row[:col], msilib.Binary(row[col]), *row[col + 1:])` but that’s Python 3.5+ I don’t know i that’s OK ?

marcelotduarte:
cx-freeze supports python 3.6 to 3.9 (so f-strings, type annotation, etc)

Cimbali:
Alright so we’re good then. The 9724621 commit should fix this.

marcelotduarte:
I tested it but I don't see the icon. Using the icon sample, I see the executable with the icon, but using your sample msi_binary_data the icon doesn't show up.
image

Cimbali:
It’s another icon that this example is setting.

marcelotduarte:
How I can check it?

Cimbali:
You can achieve the same as the other sample with the following options:

```
bdist_msi_options = {
    "data": {
        "Property": [
            ('ARPPRODUCTICON', 'IconId'),
        ],
        "Icon": [
            ('IconId', 'icon.ico'),
        ],
    },
}
```

marcelotduarte:
With 7zip I see the reference in msi
image

marcelotduarte:
Thanks!

Issue #239, opened 2017-01-27
App is displayed in non-retina resolution


adius:
When I start the app with `python3 myapp.py` it's displayed with retina resolution.
When I build it with cx_Freeze it's pixelized.

<img width="301" alt="screen shot 2017-01-27 at 19 09 12" src="https://cloud.githubusercontent.com/assets/969983/22384045/79045492-e4c4-11e6-88eb-f2ab5696f51f.png">

Left: not-retina, right: retina

The app uses only https://docs.python.org/2/library/tkinter.html.
Is there a way to fix this?

David-Luis:
I have the same issue with pyqt5, low resolution when built with cx_Freeze :(

marcelotduarte:
cx_Freeze 6.7b1 has just been released.
pip install cx-Freeze==6.7b1

marcelotduarte:
cx_Freeze 6.7 has just been released.
pip install --upgrade cx_Freeze
Assuming this has been resolved.
If you had issues please open a new issue.

PR #1031, opened 2021-04-21
Added CFBundlePackageType and NSHighResolutionCapable by default to Info.plist of Darwin bundles

cainesi:
This PR adds two items to the Info.plist file of an application Bundle:
1. CFBundlePackageType: APPL, which specifies that the bundle is an application bundle
2. NSHighResolutionCapable: True, which tells the application to run in high-resolution mode by default (which seems like a more reasonable default in 2021--without this frozen GUI apps can appear blurry).


marcelotduarte:
Does this patch fix #239?

cainesi:
Yes it does!

(I've been fixing it in my own apps via the plist_items parameter for a while. Did not see that there was an open issue for it.)

marcelotduarte:
Good!
I think that the previous title "Added CFBundlePackageType and NSHighResolutionCapable to an Darwin bundles Info.plist." should be better, no?

cx_Freeze/macdist.py
```diff
@@ -237,6 +237,12 @@ def create_plist(self):
         # Ensure CFBundleExecutable is set correctly
         contents["CFBundleExecutable"] = self.bundle_executable
 
+        # Specify that bundle is an application bundle
+        contents["CFBundlePackageType"] = "APPL"
+
+        # Cause application to run in high-resolution mode by default
+        contents["NSHighResolutionCapable"] = "True"
```
marcelotduarte: Is this value set to true but the user can change it in the plist_items?
Can you document this?

cainesi:
Yes, fine with me.

cx_Freeze/macdist.py
cainesi: Done.  I also tweaked the code so that the new default values will not be used if there is a custom_info_plist specified.

marcelotduarte:
Thanks!

PR #1032, opened 2021-04-22
Add extension registration on Windows

Cimbali:
This PR adds the ability to register a program as handling some file types on windows, using msi features, with minimal supplement of direct registry modifications.

I tried to keep the interface simple, so there’s a single option to modify, with a list of dicts, one per file-type association, which is comprised of:
1. a file type (extension without leading .)
2. an executable
3. a verb, e.g. “open”
4. optionally arguments, e.g. `"%1"` for proper quoting, but any other program arguments can be passed (see sample),
5. optionally a context menu display name for the association, defaults to `{program name} {verb}`
6. optionally a mime type associated with the extension

Most affected MSI tables are really straightforward: extension, verb, mime. 2 slightly less so:

**Component**: Executables have to be bundled into separate `Component`s (with the executable as its `keyfile`).

There was previously one automatically-created `Component` entry per directory, there now are 1 per directory + 1 per executable. If a directory has the same name as a Component we use for an executable, then a second Component with the same identifier will be created, however these have to be unique. I therefore tried to minimize the risk of Component id clashes by naming the executable Components `_cx_executable{N}_{executable base name}`, however I’m open to better suggestions. An alternative (that I find less pleasant because it relies on `msilib` internals) is to add the executable Component ids to `msilib._directory` so they won’t be reused.

**ProgId**: The `progId` are as close as possible to [recommendations](https://docs.microsoft.com/en-us/windows/win32/shell/fa-progids), which are `[Vendor or Application].[Component].[Version]`.

The 2 types of registry keys that are added are needed:
- `FriendlyAppName`: to set a display name other than the executable’s file name (`Hello Program` instead of `hello.exe` in the Open With submenu)
- `ApplicationCompany`: to have the application show directly in the « Open With » submenu rather than only in the « Choose another program » dialog from the « Open With » submenu.

----

Building & installing the sample I added modifies the context menu for the specified extensions / file types:
- it primarily adds an entry the « Open With » submenu,
- and also modifies the top options of the context menu (one per verb for the extension), when the program is selected as default handler for the file type

hello-world

The executable + argument are called with the file when an action is selected, e.g.:

action

### DIFF
--- a/cx_Freeze/windist.py
+++ b/cx_Freeze/windist.py
@@ -40,6 +40,7 @@ class bdist_msi(distutils.command.bdist_msi.bdist_msi):
         ("product-code=", None, "product code to use"),
         ("install-icon=", None, "icon path to add/remove programs "),
         ("all-users=", None, "installation for all users (or just me)"),
+        ('extensions=', None, 'Extensions for which to register Verbs'),
     ]
     x = y = 50
     width = 370
@@ -296,7 +297,21 @@ class bdist_msi(distutils.command.bdist_msi.bdist_msi):
         while todo:
             dir = todo.pop()
             for file in os.listdir(dir.absolute):
-                if os.path.isdir(os.path.join(dir.absolute, file)):
+                sep_comp = self.separate_components.get(os.path.relpath(
+                    os.path.join(dir.absolute, file),
+                    self.bdist_dir,
+                ))
+                if sep_comp is not None:
+                    restore_component = dir.component
+                    dir.start_component(
+                        component=sep_comp,
+                        flags=0,
+                        feature=f,
+                        keyfile=file,
+                    )
+                    dir.add_file(file)
+                    dir.component = restore_component
+                elif os.path.isdir(os.path.join(dir.absolute, file)):
                     newDir = msilib.Directory(
                         db,
                         cab,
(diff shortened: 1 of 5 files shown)


Cimbali:
Forgot to add, this is tested for all-users as well as current-user installs. There’s a built version of the sample [available here](https://ci.appveyor.com/project/Cimbali/cx-freeze/build/artifacts) 

To run it for all users I did  `msiexec /i "Hello Program-0.1-mingw.msi" ALLUSERS=1`

marcelotduarte:
@Cimbali I liked the description of the PR, aka the first comment. I think you should create a README.md and put this information in it. (In https://github.com/marcelotduarte/cx_Freeze/tree/main/cx_Freeze/samples/icon has a basic readme).
You can put the information of your last comment too.

marcelotduarte:
Thanks!

cx_Freeze/windist.py
```diff
@@ -296,7 +297,21 @@ def add_files(self):
         while todo:
             dir = todo.pop()
             for file in os.listdir(dir.absolute):
-                if os.path.isdir(os.path.join(dir.absolute, file)):
```
marcelotduarte, line 299: @Cimbali Apparently this change is the root cause of issue #1184 
Can you take a look?

PR #1035, opened 2021-04-25
Split up Freezer object

cainesi:
@marcelotduarte, any objections if I start work on splitting the platform-specific code into subclasses of Freezer, as we previously discussed?  (If that's okay, I'll use this branch for the work.)


marcelotduarte:
Go ahead!

marcelotduarte:
This branch has conflicts, apparently with your PR #883 merged yesterday. I would like to go testing while you do it. In fact, I think it's getting better than I imagined.

cx_Freeze/dist.py
```diff
@@ -236,7 +237,18 @@ def run(self):
...
+        if sys.platform == "win32":
+            freeze_class = WinFreezer
+        elif sys.platform == "darwin":
+            freeze_class = DarwinFreezer
+        elif sys.platform == "linux":
+            freeze_class = LinuxFreezer
+        else:
+            raise Exception(f'Unknown platform: {sys.platform}')
+
+        freezer: Freezer = freeze_class(
```
marcelotduarte: I think that this should be moved to freeze.py, so the API maintains compatibility. While this is being tested, you can leave it at that.

cainesi:
Yeah. I was working from main, so it does not include the not yet merged changes. I am still in the process of separating out the platforms, and once that is done I'm going to clean it up a lot. So this is still very preliminary, and far from ready to merge!  :)

cx_Freeze/dist.py
cainesi: Okay, that's easy enough to fix. We can make it so that the existence of the subclasses is transparent for someone using the Freezer constructor.  (just made a commit to do that)

marcelotduarte:
I am loving that you are doing this separation, but I need you to only do the separation, and then you add news for mac, otherwise, I will be confused to review and merge your work.

cainesi:
Sorry, got a bit carried away.  I'll keep this one as the pure split up, and move the subsequent Darwin and other improvements to a separate PR. 

FYI, my plan is to have the Freezer identify the files to copy, and have a separate object (FileCopier) that tracks and organizes the files to be copied (including pulling in required dependencies).  But that will all be in the post-split PR.

marcelotduarte:
I'm sorry too, but I need to check that the compatibility remains and that the change will not have any consequences for other systems.

> FYI, my plan is to have the Freezer identify the files to copy

We will discuss this later. I think it would be interesting, so as not to waste time, and I don't feel sorry and have to reject it, we talked about the ideas. At first, I'm trying to optimize on windows and linux, and I recently managed to shrink many dlls that were copied over and over into several folders.

cx_Freeze/freezer.py
```diff
@@ -505,15 +420,10 @@ def onerror(*args):
         self.binPathExcludes = [os.path.normcase(name) for name in paths]
 
         # control runtime files
+        #TODO: Consider if we can put this data only in WinFreezer
         self.runtime_files = set()
         self.runtime_files_to_dup = set()
```
marcelotduarte: we can put this data only in WinFreezer

cx_Freeze/freezer.py
```diff
@@ -41,6 +41,19 @@
...
+        if sys.platform == "win32":
+            instance =  super().__new__(WinFreezer)
+        elif sys.platform == "darwin":
+            instance =  super().__new__(DarwinFreezer)
+        elif sys.platform == "linux":
+            instance =  super().__new__(LinuxFreezer)
+        else:
+            raise Exception(f'Unknown platform: {sys.platform}')
+        return instance
+
```
marcelotduarte, line 54: Very good!

cx_Freeze/freezer.py
```diff
@@ -305,7 +305,7 @@ def _GetDependentFiles(self, path, darwinFile: DarwinFile = None) -> List:
         imagehlp library on Windows, otool on Mac OS X and ldd on Linux);
         limit this list by the exclusion lists as needed"""
         path = os.path.normcase(path)
-        dependentFiles = self.dependentFiles.get(path, [])
+        dependentFiles = self.dependentFiles.get(path, None)
```
marcelotduarte: This should be in a separate patch if it is real. I think that with the change when the same dependency is found, the ldd/otool will be run again and detect the same dependencies.

cx_Freeze/freezer.py
```diff
@@ -896,7 +892,8 @@ def _PlatformGetDependentFiles(self, path, darwinFile: DarwinFile = None) -> Lis
         return dependentFiles
 
 class LinuxFreezer(Freezer):
-    def _PlatformInit(self):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         self.patchelf = Patchelf()
         return
```
marcelotduarte, line 967: Until this patch, I think that the separation is ready to adjust, review, test.

cx_Freeze/freezer.py
cainesi: Okay.

cx_Freeze/freezer.py
cainesi: Yeah, this one was a separate issues I found when going through the code.  Previously we had "self.dependentFiles.get(path,[])" which would return the list of dependencies, if previously calculated, or [] by default.  The problem was that the code could not distinguish between cases where (i) dependencies had not previously been checked, so that [] was returned by default, and (ii) dependencies were previously checked, but no dependencies were found, so [] as stored in the dictionary.  In case (ii) you could end up running the dependency check code multiple times (unnecessarily) on the same file.  I can put this in a separate PR if you prefer.

cainesi:
> I'm sorry too, but I need to check that the compatibility remains and that the change will not have any consequences for other systems.
> 
> > FYI, my plan is to have the Freezer identify the files to copy
> 
> We will discuss this later. I think it would be interesting, so as not to waste time, and I don't feel sorry and have to reject it, we talked about the ideas. At first, I'm trying to optimize on windows and linux, and I recently managed to shrink many dlls that were copied over and over into several folders.

Sure.  Happy to discuss.  I don't think the extra (post-split-up) changes will be too much work, so not a big loss either way. I would create an experimental version first that created two copies of the frozen application in parallel, using the old and new code, so that we could check that it gave identical results.

marcelotduarte:
> so not a big loss either way

I'm not saying that I'm going to reject, just that I hope I don't have to reject.

cx_Freeze/dist.py
```diff
@@ -10,10 +10,11 @@
 import distutils.version
 import os
 import sys
+from typing import Type
 import warnings
 
 from .common import normalize_to_list
-from .freezer import Freezer
+from .freezer import Freezer, WinFreezer, DarwinFreezer, LinuxFreezer
```
marcelotduarte: This can be reverted too...

cx_Freeze/freezer.py
```diff
@@ -89,30 +103,10 @@ def __init__(
         self.zipIncludePackages = zipIncludePackages
         self.zipExcludePackages = zipExcludePackages
         self._VerifyConfiguration()
+        return
```
marcelotduarte, line 104: this 'return' is unnecessary

marcelotduarte:
I tested it in CI. Simple test. In Linux and macOS okay.
On windows:
```
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\hostedtoolcache\windows\Python\3.8.9\x64\Scripts\cxfreeze.exe\__main__.py", line 7, in <module>
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\site-packages\cx_Freeze\cli.py", line 262, in main
    freezer.Freeze()
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\site-packages\cx_Freeze\freezer.py", line 669, in Freeze
    self._CopyFile(
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\site-packages\cx_Freeze\freezer.py", line 160, in _CopyFile
    self._PostCopy(
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\site-packages\cx_Freeze\freezer.py", line 753, in _PostCopy
    for dependent_file in self._GetDependentFiles(source):
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\site-packages\cx_Freeze\freezer.py", line 312, in _GetDependentFiles
    dependentFiles = self._PlatformGetDependentFiles(path, darwinFile)
  File "c:\hostedtoolcache\windows\python\3.8.9\x64\lib\site-packages\cx_Freeze\freezer.py", line 796, in _PlatformGetDependentFiles
    return dependentFiles
UnboundLocalError: local variable 'dependentFiles' referenced before assignment
```

cx_Freeze/freezer.py
cainesi: I agree, technically. I like to put returns at the end of all functions, just for visual clarity.  Do you feel strongly?

cainesi:
> I tested it in CI. Simple test. In Linux and macOS okay.
> On windows:

Fixed.  Had just missed the line to initialize dependentFiles in that function.

I've been doing all the coding on Mac, and haven't had a chance to test anything on Windows yet...

cx_Freeze/dist.py
cainesi: Fixed.

cainesi:
The platform split-up is done, and I think it is ready for review/testing.

The next thing I will look at is cleaning-up the DarwinFreezer code. But that will involve changing darwintools.py, so I thought I would leave it for a subsequent PR.

marcelotduarte:
During the week, I will try to review. Here the first test.
```
pushd /d/a/freeze-test/freeze-test/cx_Freeze/samples/simple
+ python setup.py build_exe --excludes=tkinter --include-msvcr=true
running build_exe
  File "setup.py", line 19, in <module>
    executables=executables,
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\cx_Freeze\dist.py", line 418, in setup
    distutils.core.setup(**attrs)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\cx_Freeze\dist.py", line 281, in run
    zipExcludePackages=self.zip_exclude_packages,
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\cx_Freeze\freezer.py", line 653, in __init__
    super().__init__(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\cx_Freeze\freezer.py", line 105, in __init__
    self._VerifyConfiguration()
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\cx_Freeze\freezer.py", line 398, in _VerifyConfiguration
    self._PlatformSetRuntimeFiles()
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\cx_Freeze\freezer.py", line 797, in _PlatformSetRuntimeFiles
    self.runtime_files.update(winmsvcr.FILES)
AttributeError: 'WinFreezer' object has no attribute 'runtime_files'
Error: Process completed with exit code 1.
```

cx_Freeze/freezer.py
marcelotduarte: I did not understand the question. I do not perceive visual clarity. I've been following the pep8 and other rules and got used to them.

cx_Freeze/freezer.py
```diff
@@ -39,7 +40,20 @@
...
-class Freezer:
+class Freezer(ABC):
+    def __new__(cls, *args, **kwargs):
+        # create instance of appropriate sub-class, depending on the platform.
+        instance: Freezer
+        if sys.platform == "win32":
+            instance =  super().__new__(WinFreezer)
+        elif sys.platform == "darwin":
+            instance =  super().__new__(DarwinFreezer)
+        elif sys.platform == "linux":
```
marcelotduarte: In the current code, when it is not Windows or macOS, then it is Linux. So, let's consider it this way because those who use Unix, FreeBSD, VxWorks must be able to build their application without any problems using the rules for Linux.

cx_Freeze/freezer.py
```diff
@@ -162,76 +138,26 @@ def _CopyFile(
...
-                if fix_rpath:
-                    has_rpath = self.patchelf.get_rpath(target)
-                    rpath = ":".join([f"$ORIGIN/{r}" for r in fix_rpath])
-                    if has_rpath != rpath:
-                        self.patchelf.set_rpath(target, rpath)
+    def _CopyFileNameHook(self, source: str, normalizedTarget: str) -> str:
+        return normalizedTarget
+
+    @abstractmethod
+    def _PostCopyHook(self,
```
marcelotduarte, line 153: As you know I have been trying to transform the code little by little into pep8. So, new functions can be in pep8. In this case _post_copy_hook. By the way, why did you adopt the 'hook'? I am not against (nor for), it is just to know.

marcelotduarte:
> The next thing I will look at is cleaning-up the DarwinFreezer code. But that will involve changing darwintools.py, so I thought I would leave it for a subsequent PR.

Good. I even want to discuss something about build_exe and bdist_mac in an issue that I opened the other day.

cx_Freeze/freezer.py
cainesi: Okay, makes sense. Fixed.

cx_Freeze/freezer.py
cainesi: I was just asking if you care a lot about the "return" at the function end.  I tend to have include the return because my IDE highlights it so I clearly see where the function ends, but it is not critical if you want it gone.

cx_Freeze/freezer.py
cainesi: I tried to use "hook" where some of the platforms might have wanted to run some extra code.  It ended up not being entirely consistent after all the changes I made. Once everything else is settled, it will be easy enough to change all the names to a consistent scheme. 

I'm happy to convert all the functions in Freezer to pep8 names, while I'm at it.  I do not see any downside to doing that, at least for all the private functions.  Let me know.

cainesi:
> During the week, I will try to review. Here the first test.
> ```
> [...]
> AttributeError: 'WinFreezer' object has no attribute 'runtime_files'
> Error: Process completed with exit code 1.
> ```

In your test, I assume you had set the include_msvcr option?  If so, I found the bug and it is fixed now.

marcelotduarte:
> In your test, I assume you had set the include_msvcr option? If so, I found the bug and it is fixed now.

--include-msvcr=true.

Its works. I had an error in msys, but I think that should unrelated. And I'll test without the include-msvcr.

In Linux, I had some errors. For example in bcrypt sample:
```
/work/freeze-test/freeze-test/cx_Freeze/samples/bcrypt/build/exe.linux-x86_64-3.6 ~/work/freeze-test/freeze-test
+ TEST_NAME=test_bcrypt
+ '[' -z test_bcrypt ']'
+ [[ test_bcrypt == gui:* ]]
+ [[ test_bcrypt == svc:* ]]
+ ./test_bcrypt
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/cx_Freeze/initscripts/__startup__.py", line 81, in run
    module.run()
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/cx_Freeze/initscripts/Console.py", line 36, in run
    exec(code, m.__dict__)
  File "test_bcrypt.py", line 1, in <module>
  File "/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/bcrypt/__init__.py", line 25, in <module>
    from . import _bcrypt  # type: ignore
ImportError: libffi-806b1a9d.so.6.0.4: cannot open shared object file: No such file or directory
Error: Process completed with exit code 255.
```

cainesi:
> > In your test, I assume you had set the include_msvcr option? If so, I found the bug and it is fixed now.
> 
> --include-msvcr=true.
> 
> Its works. I had an error in msys, but I think that should unrelated. And I'll test without the include-msvcr.
> 
> In Linux, I had some errors. For example in bcrypt sample:

Without --include-msvcr the program should already have worked fine. The error was that I was defining the --include-msvcr related variables to late in the object initialization (so they could end up being referenced before assignment).

I have never used cx_freeze on linux before.  The split up was intended not to change anything about the operation on linux, so I must have messed something up.  On a very quick look, I think I probably made a mistake in LinuxFreezer._PostCopyHook.  Will take a closer look later.

cainesi:
I fixed the problem with the bcrypt sample.  As expected, it was a bug I had accidentally introduced in LinuxFreezer._PostCopyHook.

marcelotduarte:
Good. Its works! I'll make some more tests with 1 or 2 projects of mine.
Also, I'll test with some examples that I have that I know that there are problems to see if the issues remain ;-)

cx_Freeze/freezer.py
marcelotduarte: I'll remove all of them.

cx_Freeze/freezer.py
```diff
@@ -89,66 +101,28 @@ def __init__(
...
-        if norm_target_name in self.runtime_files:
-            target_name = os.path.basename(target)
-            target = os.path.join(self.targetdir, "lib", target_name)
-            # vcruntime140.dll must be in the root and in the lib directory
-            if norm_target_name in self.runtime_files_to_dup:
-                self.runtime_files_to_dup.remove(norm_target_name)
-                self._CopyFile(source, target, copyDependentFiles=False)
-                target = os.path.join(self.targetdir, target_name)
-            normalizedTarget = os.path.normcase(os.path.normpath(target))
+        normalizedTarget = self._CopyFileNameHook(source=source, normalizedTarget=normalizedTarget)
```
marcelotduarte: You should pass source and target to _CopyFileNameHook, because without this, the names are not preserved.
And inf _CopyFileNameHook
target_name = os.path.basename(target)
was changed to 
target_name = os.path.basename(normalizedTarget)
which returns with a regression resolved in 6.5
edit: do not cause the regression because here only the c runtimes are affected

cx_Freeze/freezer.py
marcelotduarte: We will do this after the PR is merged, otherwise, this patch will increase a lot.

cx_Freeze/freezer.py
```diff
@@ -267,45 +191,13 @@ def _FreezeExecutable(self, exe):
...
-        # Always copy the dynamic libraries into lib folder
-        target_dir = os.path.join(self.targetdir, "lib")
+        self._PlatformAddExtraExecutableDependencies(dependent_files=dependent_files)
+
         for source in dependent_files:
-            # but python dlls should be copied within the executable
-            if sys.platform == "win32" and source.endswith(python_libs):
-                target = os.path.join(self.targetdir, os.path.basename(source))
-            else:
-                target = os.path.join(target_dir, os.path.basename(source))
```
marcelotduarte, line 287: I see some errors when the sample are run in a venv. This behavior isn't present in WinFreezer. A _CopyTopDependency should exist in it and take care of that.

cx_Freeze/freezer.py
```diff
@@ -267,45 +191,13 @@ def _FreezeExecutable(self, exe):
...
-                    copyDependentFiles=True,
-                    includeMode=True,
-                    machOReference=cachedReference,
-                )
-            else:
-                self._CopyFile(
-                    source, target, copyDependentFiles=True, includeMode=True
-                )
+            # Store dynamic libraries in appropriate location for platform
+            target = os.path.join(self._PlatformExecutableDependencyDir(), os.path.basename(source))
```
marcelotduarte: I am in doubt if the method _PlatformExecutableDependencyDir needs to exist. You can restore the line 
target_dir = os.path.join(self.targetdir, "lib")
before the for and take care of the correct path in _CopyTopDependency in WinFreezer

cx_Freeze/freezer.py
```diff
@@ -318,157 +210,71 @@ def _FreezeExecutable(self, exe):
...
-                    else:
-                        raise
-            else:
-                target_icon = os.path.join(
-                    self.targetdir, os.path.basename(exe.icon)
-                )
-                self._CopyFile(exe.icon, target_icon, copyDependentFiles=False)
+        target_lib_dir = os.path.join(self.targetdir, "lib")
+        self._CopyIcon(exe=exe, target_dir=target_lib_dir,
+                           target_path=target_path)
```
marcelotduarte: _CopyIcon only uses the exe param (and  target_path on windows), so target_lib_dir can be removed.

marcelotduarte (changes requested):

cx_Freeze/freezer.py
cainesi: Fine with me.

cx_Freeze/freezer.py
cainesi: I think there was a bug there, since the replaced code could have changed both target and normalizedTarget.  I have revised to more accurately reflect the original behaviour (not the function will receive and return both target and normalizedTarget).

cx_Freeze/freezer.py
cainesi: Okay, makes sense.

cx_Freeze/freezer.py
cainesi: Agreed. Fixed. (Not sure what I was thinking there!)

cx_Freeze/freezer.py
marcelotduarte: It's not a bug. if 'target' can be changed, 'normalizedTarget' needs to reflects that change.

cainesi: Yes, exactly.  The bug was in my code, and I have fixed it my latest commit (0c0cfae8) so that both target and normalizedTarget will be changed.

cx_Freeze/freezer.py
cainesi: Okay, that makes sense.  I eliminated _PlatformExecutableDependencyDir and moved the code into a new _CopyTopDependency for WinFreezer instead.

cx_Freeze/freezer.py
cainesi: I think this behaviour was reproduced in WinFreezer.  Originally the line 
```
target = os.path.join(self._PlatformExecutableDependencyDir(), os.path.basename(source))
```
should have had the same effect, but the logic has now been moved into _CopyTopDependency() as you suggested below.

cx_Freeze/freezer.py
```diff
@@ -89,66 +101,28 @@ def __init__(
...
-            target_name = os.path.basename(target)
-            target = os.path.join(self.targetdir, "lib", target_name)
-            # vcruntime140.dll must be in the root and in the lib directory
-            if norm_target_name in self.runtime_files_to_dup:
-                self.runtime_files_to_dup.remove(norm_target_name)
-                self._CopyFile(source, target, copyDependentFiles=False)
-                target = os.path.join(self.targetdir, target_name)
-            normalizedTarget = os.path.normcase(os.path.normpath(target))
+        # fix the target path for C runtime files (should do nothing on Darwin...)
+        trarget, normalizedTarget = self._CopyFileNameHook(source=source, target=target, normalizedTarget=normalizedTarget)
```
marcelotduarte: Now have a typo: trarget

cainesi: Sorry, fixed. (That typo would have had no effect, anyway, though, since the function does not do anything in DarwinFreezer.)

cx_Freeze/freezer.py
```diff
@@ -845,6 +631,429 @@ def Freeze(self):
...
+        return
+
+    def _PostCopyHook(self,
+                      source,
+                      target,
+                      normalizedSource,
+                      normalizedTarget,
+                      copyDependentFiles,
+                      includeMode=False):
+        # TODO: remove usages of source/target (only use normalied paths)
```
marcelotduarte: You can remove the TODO. At least I need the raw source here, not the normalized.

cainesi: Done. Removed.

marcelotduarte:
Thanks!

PR #1042, opened 2021-05-02
Fix minor glitch in _GetDependentFiles.

cainesi:
As it currently is, the dependency-finding code could run repeatedly on any file that does not have any dependencies (because [] would be stored into the self.dependentFiles dictionary for that file, which is the same as the default value).  This PR fixes that by using None as the default.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -400,8 +400,9 @@ class Freezer:
         imagehlp library on Windows, otool on Mac OS X and ldd on Linux);
         limit this list by the exclusion lists as needed"""
         path = os.path.normcase(path)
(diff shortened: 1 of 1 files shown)


marcelotduarte:
The fix in this PR was merged with the PR #1035 
But, you can point me to a test that proves that this is correct?

cainesi:
I don't think the old code would have caused any actual errors in the existing cx_freeze, so I don't have a test that fails with it. It just wasted effort (due to repeatedly re-analyzing the same file). It did cause issues with the new Darwin code that I am working on, which is how I noticed.

marcelotduarte:
Okay. Thanks!

PR #1056, opened 2021-05-11
fade to black

marcelotduarte:
The packages pyupgrade and black are used. Black is a PEP 8 compliant opinionated formatter. 
pyupgrade --py36-plus file.py
black -l 79 -t py36 file.py


marcelotduarte:
CC: @cainesi

PR #1122, opened 2021-06-23
Small small changes to resolve code warnings

cainesi:
These a small small glitches in the code that I noticed when working through it.  Should not have any real effect on the operation of the program, but will result in future warnings in IDEs.


marcelotduarte:
Thanks!

PR #1156, opened 2021-07-17
Update PyQt5 sample to document and use new function 'get_qt_plugins_paths'

marcelotduarte:


marcelotduarte:
@cainesi I'm updating hooks for pyqt5 and then I'll do it for pyside2, trying to solve a series of open issues, for several systems. For macOS I'm just testing on CI. So if you can test it with your systems, it's not breaking anything.
As a bonus, I created a function that can be imported and makes it easy to add plugins in pyqt5. See the new pyqt5 sample. Basically, it uses:
```
from cx_Freeze.hooks import get_qt_plugins_paths
include_files = get_qt_plugins_paths("PyQt5", "platforms")  # "platforms" is used here for demonstration purposes.
```

PR #1234, opened 2021-09-03
Fixes for the existing nose tests.. bar one

TechnicalPirate:
**Problem** - The existing tests in the repository had failures due to not being updated with newer releases of `cx_freeze` this is a barrier to contributors.

**Solution** - This PR fixes the failing tests by 

* Passing an expected dependency ( `ConstantsModule` ) to the majority of failing tests
* Updates the `test_process_path_specs` to account for the fact the code under test now uses `pathlib.Path` objects
* **DISABLES** `test_FindModule_from_zip` - the under lying code will return `None` as i couldn't find any `isinstance` checks for handling loading from `.zip` - Flagging this one for @marcelotduarte :)

Related Discussion: [here](https://github.com/marcelotduarte/cx_Freeze/discussions/1186)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ __pycache__
 *.pyc
 sample.xlsx
 dump.sql
-
+# PyCharm configuration files
+.idea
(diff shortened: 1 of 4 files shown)


marcelotduarte:
Thanks!

PR #1244, opened 2021-09-13
Fix memory leaks and other types of leaks

### LINKED ISSUE #1040: Possible (tiny) Memory Leak in Win32Service.c
I noticed in Win32Service.c that there are several places where a `PyObject` reference is acquired but then the ref count is never decremented. For example, in `Service_SetupPython()` there is `module = PyImport_ImportModule(...)`, but then `Py_DECREF()` is never called. Since `PyImport_ImportModule()` returns a "new" reference, I would expect to see `Py_DECREF()` called once the function is done working with `module`, based on this doc: https://docs.python.org/3/c-api/intro.html#reference-count-details

Is there a particular reason that these reference decrements do not exist? If not, it seems to me that Win32Service.c has a very tiny memory leak; in which case I'm happy to submit a PR with the reference decrements if you wish.

 - Platform information: Windows
 - OS architecture: amd64
 - cx_Freeze version: 6.6
 - Python version: 3.8

marcelotduarte:
Fixes #1040 
@anthony-tuininga Can you review this for me?

### DIFF
--- a/source/bases/Win32Service.c
+++ b/source/bases/Win32Service.c
@@ -114,6 +114,7 @@ static int Service_Stop(udt_ServiceInfo* info)
 {
     PyThreadState *threadState;
     PyObject *result;
+    int ret = 0;
 
     // indicate that the service is being stopped
     if (Service_SetStatus(info, SERVICE_STOP_PENDING) < 0)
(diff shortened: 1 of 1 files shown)


source/bases/Win32Service.c
```diff
@@ -115,6 +115,7 @@ static int Service_Stop(udt_ServiceInfo* info)
 {
     PyThreadState *threadState;
     PyObject *result;
+    int ret = 0;
```
anthony-tuininga, line 117: Why these changes? If the stop() method could not be called or some other exception takes place, why are you marking the service as stopped anyway?

source/bases/Win32Service.c
```diff
@@ -164,27 +167,33 @@ static int Service_SessionChange(DWORD sessionId, DWORD eventType)
 {
     PyThreadState *threadState;
     PyObject *result;
+    char *error_message = NULL;
```
anthony-tuininga, line 169: Similar comments apply here. These changes also do not seem to be relevant for the concerns initially raised: namely, memory leaks.

source/bases/Win32Service.c
```diff
@@ -291,19 +301,25 @@ static int Service_SetupPython(udt_ServiceInfo *info)
 
     // acquire the __main__ module
     module = PyImport_ImportModule("__main__");
-    if (!module)
-        return LogPythonException("unable to import __main__");
+    if (!module) {
+        error_message = "unable to import __main__";
+        goto end;
```
anthony-tuininga, line 302: I am not at all a fan of goto. If necessary, you can create a separate method (Service_SetupPythonHelper or some such) which will perform the work. There is no real need, though, since if this fails, the entire executable will fail and the process will end.

source/bases/Win32Service.c
marcelotduarte: This patch I should have put separately. There is another issue that I think may be related.

source/bases/Win32Service.c
marcelotduarte: I call this "other types of leaks", because the thread is aquired but not released.

source/bases/Win32Service.c
marcelotduarte: I don't like 'goto' either. I recently added a new function in util and used 'if succeded' but it's not all good either. As I have been reading some python codes and I see that in some situations they surrender to goto, I surrendered too....

source/bases/Win32Service.c
```diff
@@ -127,32 +128,34 @@ static int Service_Stop(udt_ServiceInfo* info)
...
+    } else
+        ret = LogPythonException("unable to create new thread state");
 
     // indicate that the service has stopped
     if (Service_SetStatus(info, SERVICE_STOPPED) < 0)
-        return LogWin32Error(GetLastError(), "cannot set service as stopped");
+        ret = LogWin32Error(GetLastError(), "cannot set service as stopped");
 
     // now set the control event
     if (!SetEvent(gControlEvent))
```
deymundson-cx, line 154: The logic here is now different, since it used to exit early. I'm not familiar with `Service_SetStatus` or `SetEvent` so perhaps this is benign?

source/bases/Win32Service.c
```diff
@@ -468,45 +500,70 @@ static int Service_Install(wchar_t *name, wchar_t *configFileName)
 //-----------------------------------------------------------------------------
 static int Service_Uninstall(wchar_t *name)
```
deymundson-cx: I noticed that reference counts in `Service_Install` are not handled, but they are handled in `Service_Uninstall`. Was `Service_Install` left out intentionally? In any case, my main concern is the reference counts in  `Service_SetupPython` and `Service_Run` are addressed (which they are), since any memory leak as part of install or uninstall will be very short lived.

PR #1251, opened 2021-09-18
Apply ad-hoc signature to modified libraries

cedk:


marcelotduarte:
@cedk 

Only on this platform (arm64) this is necessary or have you only tested on it?

cedk:
For me the problem appears only on Big Sur with arm64.
Indeed I followed what Homebrew is doing.

marcelotduarte:
Thanks!

PR #1252, opened 2021-09-21
Fix #1250 missing executable on successive same-version installs 

### LINKED ISSUE #1250: Problems when installing new version of application from MSI
**Describe the bug**
When installing a new version of already installed application from MSI installer (so that the previously installed version is automatically uninstalled before the new installation), the executable files are omitted.  

Problem was introduced by 99b3ad90.

**To Reproduce**
1) Create MSI installer with cx_freeze.
2) Install from MSI
3) Create second MSI with the same executable and upgrade code.
4) Install from second MSI

**Expected behavior**
Should uninstall the first version and fully install the second.  Instead, the executable from the second installer is omitted.

**Additional context**
The problem appears to be caused by putting the executable in its own component, with the executable file as the keypath.   I think what happens on the second install is that: it first scans for the existing files and determines that the component for the executable is already there so does not need to be installed, then removes the entire existing installation, then installs the new new version except the executable component that it determined was unnecessary.

Problem goes away if the following code in windist.py is disabled.

```
if sep_comp is not None:
   restore_component = dir.component
   dir.start_component(
      component=sep_comp,
      flags=0,
      feature=f,
      keyfile=file,
   )
   dir.add_file(file)
   dir.component = restore_component
```

@cimbali -- If you are reading, was there a need to put the executables into separate components?  I am not familiar enough with the extension registration system to know if this was somehow required.

As a "quick and dirty" fix, maybe we could only put the executables into separate components if extensions are being registered?

Cimbali:
Set REINSTALLMODE to force installing same-version executables

In development, or other sitations with successive installs with
same-version executables, this avoids the following “strange” behaviour:

1. installer determines previous executable does not need replacing
2. the Upgrade table determines the previous executable is part of an
   older product
3. the RemoveExistingProducts action removes the previous installation,
   including the previous executable
4. during install, the new executable is not installed as it was deemed
   unnecessary initially and this appraisal has not been affected by
   the previous version uninstall.

### DIFF
--- a/cx_Freeze/windist.py
+++ b/cx_Freeze/windist.py
@@ -75,7 +75,7 @@ class bdist_msi(distutils.command.bdist_msi.bdist_msi):
         # This is needed in case the AlwaysInstallElevated policy is set.
         # Otherwise installation will not end up in TARGETDIR.
         msilib.add_data(
-            self.db, "Property", [("SecureCustomProperties", "TARGETDIR")]
+            self.db, "Property", [("SecureCustomProperties", "TARGETDIR;REINSTALLMODE")]
         )
         msilib.add_data(
             self.db,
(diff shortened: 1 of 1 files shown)


marcelotduarte:
Thanks!

PR #1260, opened 2021-09-30
fix bugs in certain subprocess calls

cainesi:
The argument lists already contained the name of the executable. 

FYI. Seems like the latest version might have some other problems on OSX (my frozen app is seg faulting).  Will try to look into it later.

### DIFF
--- a/cx_Freeze/macdist.py
+++ b/cx_Freeze/macdist.py
@@ -87,7 +87,7 @@ class bdist_dmg(Command):
             )
 
         # Create the dmg
-        if subprocess.call(("hdiutil", createargs)) != 0:
+        if subprocess.call(createargs) != 0:
             raise OSError("creation of the dmg failed")
 
     def run(self):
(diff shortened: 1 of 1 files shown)


marcelotduarte:
Thanks!

marcelotduarte:
> Seems like the latest version might have some other problems on OSX

Can be in the support for pathlib.Path

Running the samples in python 3.9.6 it passed. Now in CI has python 3.9.7 I get errors building wheels, so I cannot test.

cainesi:
In my case I was using Python 3.8.

cainesi:
Looks like the issue is probably related to the PyQt hooks.  App stops running on 221f164c (with: "NameError: name 'QApplication' is not defined"), then on the next commit (e4a972da) the seg-fault appears.  I'm guessing it's something to do with this new load_PyQt5() hook.  When I comment out that hook, things run fine.

I do not fully understand that new hook.  Also, how does it interact with the various existing hooks that copy plugins to other locations with copy_qt_plugins()?

I'll open a new issue for this.

PR #1278, opened 2021-11-07
Support for latest opencv-python release

### LINKED ISSUE #1275: Executables with opencv-python>=4.5.4 fail
**Describe the bug**
After building an executable which depends on the latest (4.5.4) version of opencv-python, this error ocurrs on running:

```ImportError: OpenCV loader: missing configuration file: ['config.py']. Check OpenCV installation.```

When comparing the `lib/cv2` folder produced by cx_Freeze to `lib/site-packages/cv2` in a python environment, the files `config.py` and `config-{python_version}.py` (`config-3.8.py` in my case) are missing.

From [this issue](https://github.com/opencv/opencv-python/issues/569), it seems that recent updates to opencv-python have made it more difficult for tools to infer what files are required. Adding the following to `hooks.py` makes things work, but I don't know enough about cx_Freeze to say whether this does things in the right way, or even if it works beyond my machine.

```
def load_cv2(finder: ModuleFinder, module: Module) -> None:
    """The cv2 module requires additional configuration files"""
    dest_dir = Path("lib", "cv2")
    cv2_dir = module.path[0]
    for path in cv2_dir.glob("config*.py"):
        finder.IncludeFiles(path, dest_dir / path.name)
```

**To Reproduce**
This bug is present for any application build with opencv-python>=4.5.4 as a dependency.

**Expected behavior**
I expected things to work the same as when I was using version 4.5.3

 - Platform information (e.g. Ubuntu Linux 16.04): Windows 10
 - OS architecture (e.g. amd64): amd64
 - cx_Freeze version [e.g. 6.6]: 6.8.1
 - Python version [e.g. 3.9]: 3.8

JJReyka:
Fixes #1275.

This adds a hook which includes additional .py files which are now required in the latest opencv-python release (4.5.4). Since these files are directly executed (but not imported) in `cv2/__init__.py`, they may have been missed out when cx_Freeze creates a build depending on this library.

The `config.py` and `config-{python_version}.py` files are not part of the previous release, so nothing extra should be included if building with the old library.  

When testing on Ubuntu, I also found that a folder containing shared libraries `opencv-python.libs` was not copied into builds, but will be if the package itself is included.

This has been tested on opencv-python 4.5.3/4 and Windows 10/Ubuntu 20.04. But bearing in mind the additional issue which came up on Linux, it would be a great help if anyone could confirm this works on MacOS!

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -379,6 +379,17 @@ def load__ctypes(finder: ModuleFinder, module: Module) -> None:
         finder.IncludeFiles(dll_path, Path("lib", dll_name))
 
 
+def load_cv2(finder: ModuleFinder, module: Module) -> None:
+    """
+    Versions of cv2 (opencv-python) above 4.5.3 require additional
+    configuration files.
+    """
+    dest_dir = Path("lib", "cv2")
+    cv2_dir = module.path[0]
+    for path in cv2_dir.glob("config*.py"):
+        finder.IncludeFiles(path, dest_dir / path.name)
+
+
 def load_cx_Oracle(finder: ModuleFinder, module: Module) -> None:
     """
     The cx_Oracle module implicitly imports datetime; make sure this
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -383,19 +383,17 @@ def load_cv2(finder: ModuleFinder, module: Module) -> None:
     """
     Versions of cv2 (opencv-python) above 4.5.3 require additional
     configuration files.
+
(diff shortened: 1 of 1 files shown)


marcelotduarte:
> This has been tested on opencv-python 4.5.3/4 and Windows 10/Ubuntu 20.04. But bearing in mind the additional issue which came up on Linux, it would be a great help if anyone could confirm this works on MacOS!

The additional .libs is from manylinux wheels, as explained in [readme](https://github.com/opencv/opencv-python/blob/master/README.md#ci-build-process) section 3 and 5.

Then, this should work on macOS!

marcelotduarte:
Thanks!

PR #1279, opened 2021-11-09
Include updated libffi 

JJReyka:
Just a small thing I noticed while working on #1275.
On windows, upgrading from python 3.8.11 -> 3.8.12 means a newer version of `libffi` (`libffi-8.dll` vs `libffi-7.dll`) comes with `ctypes`. This just changes the search pattern to include any version rather than 7 specifically.

I've tested this on Windows with the python versions above.

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -374,7 +374,7 @@ def load__ctypes(finder: ModuleFinder, module: Module) -> None:
     libffi dll to be present in the build directory.
     """
     if WIN32 and sys.version_info >= (3, 8) and not MINGW:
-        dll_pattern = "libffi*-.dll"
+        dll_pattern = "libffi-*.dll"
         dll_dir = Path(sys.base_prefix, "DLLs")
         for dll_path in dll_dir.glob(dll_pattern):
             finder.IncludeFiles(dll_path, Path("lib", dll_path.name))
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -374,7 +374,7 @@ def load__ctypes(finder: ModuleFinder, module: Module) -> None:
     libffi dll to be present in the build directory.
     """
     if WIN32 and sys.version_info >= (3, 8) and not MINGW:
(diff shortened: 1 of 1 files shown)


cx_Freeze/hooks.py
```diff
@@ -370,13 +370,14 @@ def pycryptodome_filename(dir_comps, filename):
...
-    In Windows, the _ctypes module in Python 3.8+ requires an additional dll
-    libffi-7.dll to be present in the build directory.
+    In Windows, the _ctypes module in Python 3.8+ requires an additional
+    libffi dll to be present in the build directory.
     """
     if WIN32 and sys.version_info >= (3, 8) and not MINGW:
-        dll_name = "libffi-7.dll"
-        dll_path = Path(sys.base_prefix, "DLLs", dll_name)
-        finder.IncludeFiles(dll_path, Path("lib", dll_name))
+        dll_pattern = "libffi*-.dll"
```
marcelotduarte: Would not be `dll_pattern = "libffi-*.dll"`?

marcelotduarte (changes requested):

cx_Freeze/hooks.py
JJReyka: You're right, well spotted! That'll teach me to try and make last second changes..

marcelotduarte:
@JJReyka From python.org and from Github Actions CI only 3.8.10 is available. [Python 3.8.12](https://www.python.org/downloads/release/python-3812/) does not have binaries. How did you test it?

marcelotduarte (approved):

JJReyka:
> @JJReyka From python.org and from Github Actions CI only 3.8.10 is available. [Python 3.8.12](https://www.python.org/downloads/release/python-3812/) does not have binaries. How did you test it?

Ah apologies for not getting back to you on this - I was testing the changes in a miniconda python environment built with `conda create -n <env name> python==3.8.12`. Unfortunately this was just on a local machine, rather than something more easily reproducible.

marcelotduarte:
Thanks!

PR #1282, opened 2021-11-15
add sentry-sdk modules

chssn:


cx_Freeze/hooks.py
```diff
@@ -77,6 +77,7 @@ def initialize(finder: ModuleFinder) -> None:
     finder.ExcludeModule("Queue")
     finder.ExcludeModule("rexec")
     finder.ExcludeModule("robotparser")
+    finder.ExcluedModule("sentry_sdk")
```
marcelotduarte: Is this entry required? Usually, only load_sentry will be needed. Entries in this section are used to delete old modules like the ones that changed their name in the transition from py2 to py3.

cx_Freeze/hooks.py
```diff
@@ -1157,7 +1158,21 @@ def load_reportlab(finder: ModuleFinder, module: Module) -> None:
     """
     finder.IncludeModule("reportlab.rl_settings")
 
-
+    
+def load_sentry(finder: ModuleFinder, module: Module) -> None:
+    """
+    The Sentry.io SDK
+    """
+    find.IncludeModule("sentry_sdk.integrations.stdlib")
```
marcelotduarte: typo (finder is the correct).

marcelotduarte (changes requested):

cx_Freeze/hooks.py
chssn: Probably not. I was trying to port across some code from PyInstaller to get the sentry_sdk working. Happy to go with those changes.

marcelotduarte (approved):

marcelotduarte:
Thanks!

PR #1303, opened 2021-12-02
Python 3.10 on windows

egormartiniuc:
For some modules "arg" argument greater than "code.co_consts" length


marcelotduarte:
Thanks for your patch. I already had a patch to this issue. My patch doesn't ignore the wrong value but gets the correct value.
So I applied PR #1306
For reference: https://bugs.python.org/issue45757

PR #1368, opened 2022-02-01
Add CI with a pre-commit file

### LINKED ISSUE #1363: Idea: Add some CI?
Hi everybody!

I think you already know GitHub Actions (GHA), because we already have a `build-wheel` workflow. However, what about using CI for linting or running the tests living in `tests/`?

This can be really useful, for example, on PRs. We can run tests on any Python we support, and (almost) every platform supported. We can also start with automation tools such as [Nox](https://nox.thea.codes) or [tox](https://tox.wiki), so both users and CI can run exactly the same.

If you're interested, and need some help, please tell me :)

DiddiLeija:
Closes #1363. This PR adds a `pre-commit` file, and a CI (`ci.yml`) file to run it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,7 +38,6 @@ repos:
   rev: 4.0.1
   hooks:
   - id: flake8
-    additional_dependencies: *flake8-dependencies
 
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.920
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,6 @@ repos:
   - id: debug-statements
   - id: end-of-file-fixer
   - id: mixed-line-ending
-  - id: requirements-txt-fixer
   - id: trailing-whitespace
 
 - repo: https://github.com/PyCQA/isort
@@ -34,13 +33,16 @@ repos:
   - id: pyupgrade
     args: [--py36-plus]
 
-- repo: https://github.com/pycqa/flake8
+- repo: https://github.com/PyCQA/flake8
   rev: 4.0.1
   hooks:
   - id: flake8
 
-- repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.920
+- repo: https://github.com/PyCQA/pylint
+  rev: v2.12.2
   hooks:
-  - id: mypy
-    args: [--show-error-codes]
+    - id: pylint
+      name: pylint
+      entry: pylint
+      language: system
+      types: [python]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
   rev: v2.29.1
   hooks:
   - id: pyupgrade
-    args: [--py37-plus]
+    args: [--py36-plus]
 
 - repo: https://github.com/pycqa/flake8
   rev: 4.0.1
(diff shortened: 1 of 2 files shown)


DiddiLeija:
@marcelotduarte I think I'm going to need permissions to run (and verify) the new CI.

marcelotduarte:
@DiddiLeija The tests failed.

DiddiLeija:
Ok, let me get that fixed... Also, I will fix anything that `pre-commit` would point.

DiddiLeija:
Ok... `pre-commit` works now. However, it started to fail on stuff (see https://github.com/marcelotduarte/cx_Freeze/runs/5024998729?check_suite_focus=true). Some of them are valid, but some others (like those generated by `tests/samples/invalid_syntax.py`) are expected, right?

So, here comes a question: where should we run `pre-commit`?

marcelotduarte:
> are expected, right?

Yes, it is. Check pyproject.toml

> where should we run `pre-commit`?

on PR?

Is there a way to avoid running in requirements for example?

DiddiLeija:
> on PR?

Sure, what I wanted to say is "on what files should we run the checks?".

> Is there a way to avoid running in requirements for example?

Sure. For the requirement files, I only have to remove one of the checks. In other files, I'm sure I can just add certain options to avoid some files.

marcelotduarte:
I think it can check py files. So, cfg, coverage, requirements, etc should be ignored.
If mypy test does not work, remove it for now, because I never used it yet.
I use in cx_Freeze pylint, black, pyupgrade (as configured in pyproject.toml, and others).

DiddiLeija:
I tried again. I removed `mypy` and `requirements-txt-fixer`. I also tried to integrate `pylint` as described [here](https://pylint.pycqa.org/en/latest/user_guide/pre-commit-integration.html). I never used `pylint` before, so tell me if you want some arguments there!

marcelotduarte:
I use the following configuration for black:
[tool.black]
line-length = 79
target-version = ["py36"]
exclude = "tests/samples/invalid_syntax.py"

marcelotduarte:
@DiddiLeija The error persists.
```
black....................................................................Failed
- hook id: black
- exit code: 2

Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.

Error: No such option: --line-length  Did you mean --line-length?
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,51 @@
+repos:
+- repo: https://github.com/psf/black
+  rev: 21.12b0
```
marcelotduarte, line 3: I use black 22.1.0
https://github.com/pypa/wheel/blob/main/.pre-commit-config.yaml use the same version and the option without "" and without space.

DiddiLeija: Thanks @marcelotduarte. I've been out of this PR for a few days, but I think today I will be experimenting locally to fix the issue. Thanks for the reminder! :)

marcelotduarte: ok, I'll fix it.

marcelotduarte:
I disabled the flake8 and pylint and fixed the configuration.
Thanks!

DiddiLeija:
Hi @marcelotduarte, thanks for pushing this. I didn't go back here due to some problems out of the internet. I'll work on tests soon :)

PR #1371, opened 2022-02-04
Added additional hooks for the Qt sqldrivers and styles plugins

cainesi:
Added additional hooks for the sqldrivers and styles plugins for PyQt5.


marcelotduarte:
Thanks!

PR #1381, opened 2022-02-14
Introduce tests in the GitHub CI

DiddiLeija:
Toward https://github.com/marcelotduarte/cx_Freeze/pull/1368#issuecomment-1031961202.

The tests will run on every Python / OS combination we support (should we do that?), so everything will be covered. They run the steps described in https://github.com/marcelotduarte/cx_Freeze/blob/main/tests/Readme.md [^1]. I'm not using any additional automation tool [^2].

cc @marcelotduarte

[^1]: Initially, I'm just running `python -m pytest`. But the tests README also suggests using `python -m pytest -m "not long" --cov="cx_Freeze" --cov-report=html`. What should we use?

[^2]: I was thinking about using Nox to automate this, but then I noticed running pytest directly was cheaper :sweat_smile:. Just ignore that by now.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,16 +10,13 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.9
     - name: Install pip dependencies
       run: |
         python -m pip install --upgrade pip setuptools
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,28 @@
+name: cx_Freeze tests
+
+on:
+  push:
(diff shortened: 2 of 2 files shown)


DiddiLeija:
> The tests will run on every Python / OS combination we support

Thinking a bit more about this, maybe we should only run the linters (pre-commit) in one Python, am I wrong?

marcelotduarte:
Tests must cover all supported versions, linters can be in only one version (py38 or py39), since black and pyupgrade, for example, accept the minimum version specification.

I fixed the test which got corrupted by isort, so merge main the tests will pass.

marcelotduarte:
@DiddiLeija Can you merge main to see the tests passing?

> 1. But the tests README also suggests using `python -m pytest -m "not long" --cov="cx_Freeze" --cov-report=html`. What should we use?

In the next step, we can use this test, if we can show the results in a comment.

DiddiLeija:
> In the next step, we can use this test, if we can show the results in a comment.

Let me investigate about doing that.

marcelotduarte:
> Let me investigate about doing that.

ok, in a separate patch.

marcelotduarte:
Thanks!

marcelotduarte:
> > In the next step, we can use this test, if we can show the results in a comment.
> 
> Let me investigate about doing that.

I found a way to do this: https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/
but as I use codecov, I don't see the need.

Issue #1431, opened 2022-03-31
failed to resolve path [@rpath/libpyside6.abi3.6.2.dylib]


emign:
**Describe the bug**
```
ERROR: Could not resolve RPath [@rpath/libpyside6.abi3.6.2.dylib] in file [/Users/username/PycharmProjects/spendenlauf/venv/lib/python3.9/site-packages/PySide6/QtNetwork.abi3.so], and could not find any likely intended reference.
[0] File: /Users/username/PycharmProjects/spendenlauf/venv/lib/python3.9/site-packages/PySide6/QtNetwork.abi3.so
  Commands:
    <MachOCommand (Load command 0 / cmd LC_SEGMENT_64)>
    <MachOCommand (Load command 1 / cmd LC_SEGMENT_64)>
    <MachOCommand (Load command 2 / cmd LC_SEGMENT_64)>
    <MachOCommand (Load command 3 / cmd LC_SEGMENT_64)>
    <MachOCommand (Load command 4 / cmd LC_DYLD_INFO_ONLY)>
    <MachOCommand (Load command 5 / cmd LC_SYMTAB)>
    <MachOCommand (Load command 6 / cmd LC_DYSYMTAB)>
    <MachOCommand (Load command 7 / cmd LC_UUID)>
    <MachOCommand (Load command 8 / cmd LC_BUILD_VERSION)>
    <MachOCommand (Load command 9 / cmd LC_SOURCE_VERSION)>
    <LoadCommand path='@rpath/libpyside6.abi3.6.2.dylib'>
    <LoadCommand path='@rpath/libshiboken6.abi3.6.2.dylib'>
    <LoadCommand path='@rpath/QtNetwork.framework/Versions/A/QtNetwork'>
    <LoadCommand path='@rpath/QtCore.framework/Versions/A/QtCore'>
    <LoadCommand path='/System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration'>
    <LoadCommand path='/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit'>
    <LoadCommand path='/usr/lib/libc++.1.dylib'>
    <LoadCommand path='/usr/lib/libSystem.B.dylib'>
    <MachOCommand (Load command 18 / cmd LC_FUNCTION_STARTS)>
    <MachOCommand (Load command 19 / cmd LC_DATA_IN_CODE)>
[...]

PR #1517, opened 2022-06-17
darwintools: Partial fix for finalizeReferences

marcelotduarte:
Partial fix for #1431

### DIFF
--- a/cx_Freeze/darwintools.py
+++ b/cx_Freeze/darwintools.py
@@ -584,7 +584,7 @@ class DarwinFileTracker:
         copied DarwinFiles and attempts to clean up any remaining references
         that are not already marked as copied. It covers two cases where the
         reference might not be marked as copied:
-        1) Files where _CopyFile was called without copyDependentFiles=True
+        1) Files where _copy_file was called without copy_dependent_files=True
            (in which the information would not have been added to the
             references at that time).
         2) Files with broken @rpath references. We try to fix that up here by
@@ -620,11 +620,7 @@ class DarwinFileTracker:
                                 "not find any likely intended reference."
                             )
                             copied_file.printFileInformation()
-                            raise DarwinException(
-                                f"finalizeReferences() failed to resolve path "
-                                f"[{reference.raw_path}] in file "
-                                f"[{copied_file.path}]."
-                            )
+                            potential_target = DarwinFile(reference.raw_path)
                         print(
                             f"WARNING: In file [{copied_file.path}]"
                             f" guessing that {reference.raw_path} "


marcelotduarte:
@cainesi I got (apparently) issue [1431](https://github.com/marcelotduarte/cx_Freeze/issues/1431#issuecomment-1146748521) to work, but I think something else can be done. As you are the original author of this code, can you review it for me and suggest what to improve?

cainesi:
Hi @marcelotduarte . Sorry it's been a while.  Hope you're doing well.

I am 99% sure that  @emign's problem is due to this line from resolveLoader():

`            return self.path.parent / path.replace("@loader_path/", "", 1)`

The problem is that it assumes that a loader_path will start with "@loader_path/", but it might just start with "@loader_path" (without the "/")--and in that case it will not be correctly handled.  So in [emign](https://github.com/emign)'s example, that is why there were 2 rpath load commands (one of which was just the path "@loader_path"), but only one calculated rpath.  Looks like there is a similar glitch in resolveExecutable() and resolveRPath().  Also, the library that it failed to find was in the @loader_path.

I think PR #1521 should resolve this (but not able to test at the moment).

The exception in finalizeReferences() should probably stay in.  If the code cannot identify the linked library at that point then something is fundamentally wrong, and it would be a pure fluke if the frozen application worked.

cainesi:
I guess this pull request can be deleted now, since a different fix was made?

marcelotduarte:
I'm still keeping it open because there is a problem to be solved, which is the premature abortion of the freeze. I don't agree with them, except for a misconfiguration.

And I just found out that in pyside6, both in Linux and in macOS, a file is missing. On Linux it freezes and runs, on macOS it raises a DarwinException. So I used a bin_includes in the sample, for both cases, but I would like that in macOS there was a 'warning' like in Linux, just informing the lack of the file.

https://github.com/marcelotduarte/cx_Freeze/blob/ba265512d82f22b1eb1578caa6d4736d913ed7c6/samples/pyside6-simplebrowser/setup.py#L21

cainesi:
The freeze process for pyside6-simplebrowser finishes for me (with those bin_excludes), but the frozen app then fails to run with a message "Could not find QtWebEngineProcess".   I think we have seen that before, and the issue is that QtWebEngineProcess is a binary file that is used by QtWebengine, but because it is launched as a new process rather than dynamically linked, the link tracing code in cx_freeze cannot pick it up.  I think the way to fix that may be to not include PySide6 in the zip file.

About "libqpdf.so" and "libqpdf.dylib", I think I have looked at that before, and the issue (if I'm remembering correctly) is that those files get pulled in via the hook for a plugin, but they are never actually dynamically loaded so the missing dependencies never get a chance to cause a problem by failing to load.  (from what I could tell, their dependencies were just missing, so they could never have been loaded by anything, whether frozen or not.) So I guess that is one example where the warning approach is better than a hard failure!

marcelotduarte:
Excuse me, I cited incorrectly. pyside6-simplebrowser for macOS is an open issue, like you described. Can be used within zip options, but the hook need a work to extract the QtWebEngineProcess from the framework...

The correct sample is pyside6:
https://github.com/marcelotduarte/cx_Freeze/blob/ba265512d82f22b1eb1578caa6d4736d913ed7c6/samples/pyside6/setup.py#L21

And you are correct about libqpdf:

> (from what I could tell, their dependencies were just missing, so they could never have been loaded by anything, whether frozen or not.)

> So I guess that is one example where the warning approach is better than a hard failure!

And yes, that's what I mean.

cainesi:
The freeze of the pyside6 sample completes for me.  There are some warnings about needing to guess how certain @rpaths should resolve, nothing terminates the build.

I have created a patch that replaces many of the fatal DarwinExceptions with a warning message.   It's in PR #1593 if you are interested.  (Recognizing that sometimes this is going to let cx_freeze spit out a frozen application that is non-functional.)

marcelotduarte:
> The freeze of the pyside6 sample completes for me.

It completes with the bin_excludes. Without bin_includes it completes in Linux e raises an error in macOS.

cainesi:
Sorry, understand now.  I see the same thing with the pyside6 sample.  Just tested it with PR #1593, and the freeze completes successfully (but with a warning) and the frozen app bundle launches.

PR #1522, opened 2022-06-19
the path of tcl and tk changed after Python3.9.6

xiaoxstz:
The change happens between Python3.9.6 and Python3.9.9

(1)The folder `tcl8.6` and `tk8.6` are moved to  `{PythonPath}\Library\lib\`  from  `{PythonPath}\` 
(2)The dll files `tcl86t.dll` and `tk86t.dll` are moved to `{PythonPath}\Library\bin\` from  `{PythonPath}\Dlls\`

### DIFF
--- a/cx_Freeze/hooks.py
+++ b/cx_Freeze/hooks.py
@@ -14,6 +14,11 @@ from .common import code_object_replace
 from .finder import ModuleFinder
 from .module import Module
 
+# get current python version
+from packaging.version import Version
+import platform
+current_version = Version(platform.python_version())
+
 MINGW = sysconfig.get_platform().startswith("mingw")
 WIN32 = sys.platform == "win32"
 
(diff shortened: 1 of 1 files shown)


marcelotduarte:
@xiaoxstz I checked with py3.9.10 that the folders are as before, but you didn't make a prediction about that. However, I imagine you are using miniconda/conda-forge, because there is this difference in folders generally. If you look at the load_sqlite3 code and do something similar for tcl/tk it will work for both cases, without needing to check versions.

marcelotduarte:
I made a PR to solve issues with Tkinter ([read the comment](https://github.com/marcelotduarte/cx_Freeze/pull/1526#issue-1284292919)).
Can you test?
To install the latest development build:

`pip install --pre --extra-index-url https://marcelotduarte.github.io/packages/ cx_Freeze`

PR #1572, opened 2022-08-01
Call InitializePython from Service_Main instead of wmain.

### LINKED ISSUE #1571: No way to use asyncio in a Windows service
The ProactorEventLoop calls `signal.set_wakeup_fd` when instantiated. This leads to the "set_wakeup_fd only works in main thread of the main interpreter" error. It seems that neither the service handler's `__init__`, `initialize`, or `run` methods are called from the main interpreter thread. Even trying to instantiate the loop at the global level doesn't work, so I guess the module is not imported from the main thread.

So there is no way to use asyncio in a Windows service ?

Platform: Windows 10, 64 bits
cx_Freeze version 6.11.1
Python 3.10.5

fraca7:
This makes the Service_Main thread the main interpreter thread, allowing
code that can only work in the main interpreter thread to be used in
a Windows service (for instance creating a ProactorEventLoop).

Also add a service sample using asyncio.

### DIFF
--- a/samples/service-asyncio/Config.py
+++ b/samples/service-asyncio/Config.py
@@ -1,0 +1,45 @@
+# ------------------------------------------------------------------------------
+# Config.py
+#   This file defines information about the service. The first four
+# attributes are expected to be defined and if they are not an exception will
+# be thrown when attempting to create the service:
+#
+#   NAME
+#       the name to call the service with one %s place holder that will be used
+#       to identify the service further.
+#
+#   DISPLAY_NAME
+#       the value to use as the display name for the service with one %s place
+#       holder that will be used to identify the service further.
+#
+#   MODULE_NAME
+#       the name of the module implementing the service.
+#
+#   CLASS_NAME
+#       the name of the class within the module implementing the service. This
+#       class should accept no parameters in the constructor. It should have a
+#       method called 'initialize' which will accept the configuration file
+#       name. It should also have a method called 'run' which will be called
+#       with no parameters when the service is started. It should also have a
+#       method called 'stop' which will be called with no parameters when the
+#       service is stopped using the service control GUI.
+#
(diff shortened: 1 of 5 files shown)


fraca7:
Fixes  #1571

marcelotduarte:
Can you rebase to get a warning in the log?

marcelotduarte:
The warning is:
```
    source/bases/Win32Service.c: In function 'Service_Main':
    source/bases/Win32Service.c:578:32: warning: passing argument 2 of 'InitializePython' from incompatible pointer type [-Wincompatible-pointer-types]
      578 |     if (InitializePython(argc, argv) < 0)
          |                                ^~~~
          |                                |
          |                                char **
    In file included from source/bases/Win32Service.c:82:
    source/bases/common.c:126:49: note: expected 'wchar_t **' {aka 'short unsigned int **'} but argument is of type 'char **'
      126 | static int InitializePython(int argc, wchar_t **argv)
          |                                       ~~~~~~~~~~^~~~
    source/bases/Win32Service.c:579:16: warning: 'return' with a value, in function returning void [-Wreturn-type]
      579 |         return 1;
          |                ^
    source/bases/Win32Service.c:573:20: note: declared here
      573 | static void WINAPI Service_Main(int argc, char **argv)
          |                    ^~~~~~~~~~~~
```

fraca7:
The simplest way to make this work would be to use directly SERVICE_TABLE_ENTRYW and StartServiceCtrlDispatcherW (Unicode versions) so that Service_Main actually gets a wchar_t** argv. Is that OK with you?

marcelotduarte:
OK

anthony-tuininga (approved):

marcelotduarte (approved):

marcelotduarte:
Thanks!

PR #1583, opened 2022-08-12
Add FAQ item for big installations (#1570)

### LINKED ISSUE #1570: cx_freeze includes too many packages in the frozen version
**Describe the bug**
The lib folder produced while freezing contains way more packages than what would be expected based on the frozen code. 

**To Reproduce**
I used the pyside2 sample from the cx_freeze repository, modified as follows to include numpy too so it becomes more representative of my real project in terms of dependencies. The setup.py script remains untouched. 

```
"""A simple script to demonstrate PySide2."""

import sys

from PySide2.QtCore import Qt
from PySide2.QtWidgets import QApplication, QLabel, QWidget
from numpy import zeros  # added line


def main():
    app = QApplication(sys.argv)
    window = QWidget()
    window.setWindowTitle("Simple")
    window.setGeometry(300, 300, 300, 300)
    label = QLabel(window)
    label.setText("Hello World!")
    label.setGeometry(0, 0, 300, 300)
    label.setAlignment(Qt.AlignCenter)
    matrix = zeros((10, 10))  # added line
    window.show()
    return app.exec_()


if __name__ == "__main__":
    sys.exit(main())
```

**Expected behavior**
The frozen lib directory should not contain packages which are not used by the frozen app, such as, in this case, pandas, torch, sqlite3, PyQt5, pythonnet... With the simple example above, the frozen app directory takes 2.6 GB!

**Screenshots**
NA

**Desktop:**
 - Platform information: Windows 10 64-bit
 - OS architecture: amd64
 - cx_Freeze version: 6.10, as provided by the WinPython 3.10.4.0 distribution
 - Python version: 3.10.4, as provided by the WinPython 3.10.4.0 distribution 

**Additional context**
I have only seen this issue so far on Python installs where a lot of packages are installed. It does not seem to be Windows-specific, as I noticed a similar behavior on a Linux install with ~200 packages installed.  
It seems to me that cx_freeze is going too far when looking for potential dependencies. A workaround could be to work with cx_freeze only when in a virtualenv, where only the necessary packages are installed (which is not the case here), but in this case I think this should be documented.  
Note that the frozen app works perfectly.

gbaudic:
This aims to improve the FAQ by adding to it the explanation given for issue #1570 .

### DIFF
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -20,15 +20,15 @@ they're dynamically loaded - e.g. by a plugin system - you have to tell
 Problems with freezing programs
 -------------------------------
 
-To determine which packages need to be copied with your application, 
-**cx_Freeze** follows the imports. If your installation contains a lot of 
-packages, this may lead to undesired behavior, such as **cx_Freeze** 
+To determine which packages need to be copied with your application,
+**cx_Freeze** follows the imports. If your installation contains a lot of
+packages, this may lead to undesired behavior, such as **cx_Freeze**
 encountering a recursion error when trying to compute the list of dependencies,
-or the `lib` folder of the frozen application containing many unnecessary 
-packages. 
-In this case, you may want to use **cx_Freeze** in a virtualenv. Alternatively, the 
-`setup_script` also offers the :option:`excludes` option to explicitly exclude 
-dependencies which would otherwise be included. 
+or the `lib` folder of the frozen application containing many unnecessary
+packages.
+In this case, you may want to use **cx_Freeze** in a virtualenv. Alternatively, the
+`setup_script` also offers the :option:`excludes` option to explicitly exclude
+dependencies which would otherwise be included.
 
 Specifying modules and packages
 -------------------------------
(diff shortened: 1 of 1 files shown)


marcelotduarte:
setuptools 64 broke the tests. I limited the version temporarily.
Can you rebase?

doc/src/faq.rst
```diff
@@ -17,6 +17,19 @@ they're dynamically loaded - e.g. by a plugin system - you have to tell
   normally be excluded (a common use is to include "libffi.so").
 * Data files are a bit more complex - see :ref:`data_files`.
 
+Problems with freezing programs
+-------------------------------
+
+To determine which packages need to be copied with your application,
+**cx_Freeze** follows the imports. If your installation contains a lot of
+packages, this may lead to undesired behavior, such as **cx_Freeze**
+encountering a recursion error when trying to compute the list of dependencies,
```
marcelotduarte, line 26: > this may lead to undesired behavior, such as **cx_Freeze**
encountering a recursion error when trying to compute the list of dependencies,

Did this happen to you or did you assume it could happen?

gbaudic:
Yes, it happened to me on a Python 3.8 install on Linux with nearly 200 
packages installed and without using a virtualenv. I did not keep the 
full trace, nor tried further debugging to spot exactly which package on 
the install triggered this behavior.

Le 13/08/2022 à 12:37, Marcelo Duarte a écrit :

anthony-tuininga (approved):

marcelotduarte:
Thanks!

PR #1591, opened 2022-08-20
Fix the filename of .msi file generated by bdist_msi.

### LINKED ISSUE #1592: Potentially incorrect filename for MSI file
**Describe the bug**
When the name of the msi file is specified by passing the target-name parameter to the bdist_msi command, the msi is created with the wrong file name. 

**To Reproduce**
In the "simple" sample, execute bdist_msi with --target-name a.msi.

An msi file is created named "a.msi-0.1-win64.msi".

**Expected behavior**
Create an msi file named "a.msi".  That has historically been the behavior.

Python 3.9 / Cx_freeze 6.11.1 / Windows 10

**Additional context**
I suspect this may have been an unintended change made when the msi module was integrated into cx_freeze?

A "quick and dirty" fix is proposed in PR #1591.

cainesi:
Fix so that if a specific .msi filename is provided by the --target-name option on the bdist_msi command, then that name will be used without alternation.

### DIFF
--- a/cx_Freeze/command/bdist_msi.py
+++ b/cx_Freeze/command/bdist_msi.py
@@ -1010,7 +1010,7 @@ class BdistMSI(bdist_msi):
         self.mkpath(self.dist_dir)
         platform = get_platform().replace("win-amd64", "win64")
 
-        msi_name:str
+        msi_name: str
         if os.path.splitext(self.target_name)[1].lower() == ".msi":
             msi_name = self.target_name
         else:
--- a/cx_Freeze/command/bdist_msi.py
+++ b/cx_Freeze/command/bdist_msi.py
@@ -1009,11 +1009,16 @@ class BdistMSI(bdist_msi):
         # make msi (by default in dist directory)
         self.mkpath(self.dist_dir)
         platform = get_platform().replace("win-amd64", "win64")
-        if self.target_version:
-            base_name = f"{self.fullname}-{platform}.msi"
+
+        msi_name:str
+        if os.path.splitext(self.target_name)[1].lower() == ".msi":
+            msi_name = self.target_name
         else:
-            base_name = f"{self.target_name}-{platform}.msi"
-        installer_name = os.path.join(self.dist_dir, base_name)
+            if self.target_version:
+                msi_name = f"{self.fullname}-{platform}.msi"
+            else:
+                msi_name = f"{self.target_name}-{platform}.msi"
+        installer_name = os.path.join(self.dist_dir, msi_name)
         installer_name = os.path.abspath(installer_name)
         if os.path.exists(installer_name):
             os.unlink(installer_name)


cx_Freeze/command/bdist_msi.py
```diff
@@ -891,6 +892,12 @@ def finalize_options(self):
         self.install_script_key = None
 
         # cx_Freeze specific
+        self.msi_name: Optional[str] = None
+        if self.target_name is not None:
```
marcelotduarte: You could create an 'else' next to 'if self.target_name is None' and maybe avoid this new variable msi_name.

marcelotduarte (changes requested):

cx_Freeze/command/bdist_msi.py
cainesi: I thought about that, but I could not see a way to do it without making other changes.  The problem is that once you get down to the run() function, you would need to know whether self.target_name was a specific name provided by the user (to be used unchanged) or something cx_freeze is generating (in which case we apparently want to add the platform name).

(I also think the variable name "msi_name" is a bit more descriptive of what is going on there. :) )

marcelotduarte: Then, I think that a few lines down, at:
self.fullname = f"{name}-{version}"
can be changed to 
```
if os.path.splitext(name)[1].lower() == ".msi":
    self.fullname = name
else:
    self.fullname = f"{name}-{version}"
```

cx_Freeze/command/bdist_msi.py
```diff
@@ -1009,11 +1016,14 @@ def run(self):
         # make msi (by default in dist directory)
         self.mkpath(self.dist_dir)
         platform = get_platform().replace("win-amd64", "win64")
-        if self.target_version:
-            base_name = f"{self.fullname}-{platform}.msi"
+        if self.msi_name is not None:
```
marcelotduarte: And here:
```
if os.path.splitext(self.target_name)[1].lower() == ".msi":
    msi_name = self.target_name
else:
    # the previous if (with base_name changed to msi_name) to be more descriptive.
```

cx_Freeze/command/bdist_msi.py
marcelotduarte: fullname is common variable used in the build/bdist commands.

marcelotduarte (changes requested):

cainesi:
Looking at the code, I'm not sure I 100% under stand how it works, and don't want to break anything.

Right now, it looks like target_name is used for a two things that are not really necessarily (naming the msi file, the product name in the msi database)?  Could those be separated?

Is there a reason we could not put all the code for determining the msi name in one place? Instead of having is spread over finalize_options() and run()?  I particularly think having two widely separated uses of os.path.splitext(string)[1].lower() == ".msi" is going to lead to confusion.

Also, does target_version do anything other than changing the name of the msi file?

cx_Freeze/command/bdist_msi.py
cainesi: That works. I made that change.

marcelotduarte (approved):

marcelotduarte:
Thanks!

marcelotduarte:
@cainesi Can you add a test?
In this file: https://github.com/marcelotduarte/cx_Freeze/blob/main/tests/test_command_bdist_msi.py#L46

cainesi:
Sure. Will do.

cainesi:
#1648 adds a test.

PR #1600, opened 2022-08-31
hooks: Support for pythonnet 3.0

xhuang2slb:
correct Python.Runtime.dll path for the pythonnet package (>3.0.0rc4).

### DIFF
--- a/cx_Freeze/hooks/__init__.py
+++ b/cx_Freeze/hooks/__init__.py
@@ -96,7 +96,7 @@ def load_clr(finder: ModuleFinder, module: Module) -> None:
     dll_name = "Python.Runtime.dll"
     finder.include_files(
         module.file.parent / "pythonnet/runtime" / dll_name,
-        Path("lib", dll_name)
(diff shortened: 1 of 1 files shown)


cx_Freeze/hooks/__init__.py
```diff
@@ -94,7 +94,10 @@ def load_clr(finder: ModuleFinder, module: Module) -> None:
     """The pythonnet package (imported as 'clr') needs Python.Runtime.dll
     in runtime."""
     dll_name = "Python.Runtime.dll"
-    finder.include_files(module.file.parent / dll_name, Path("lib", dll_name))
+    finder.include_files(
```
marcelotduarte: We need to keep supporting the previous version, so you should check if the file exists in this new path or in the old one.

marcelotduarte (changes requested):

marcelotduarte (approved):

marcelotduarte:
Thanks!

PR #1614, opened 2022-09-11
Improved documentation of initial_target_dir option on bdist_msi

cainesi:
Expand the documentation for the initial_target_dir option on bdist_msi, in order to explain how the user can achieve the "normal" result of having the application installed under "Program Files".

Instead of this, we could consider modifying bdist_msi so that [ProgramFiles64Folder] or [ProgramFilesFolder] (as applicable) as added in front of any value that the user supplies.  That would be easier for the user most of the time, but allow less flexiblity.

### DIFF
--- a/doc/src/setup_script.rst
+++ b/doc/src/setup_script.rst
@@ -349,7 +349,10 @@ command:
        installation
    * - .. option:: initial_target_dir
      - defines the initial target directory supplied to the user during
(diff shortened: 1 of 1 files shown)


marcelotduarte (approved):

marcelotduarte:
Thanks!

PR #1659, opened 2022-10-23
startup: limit the PATH in all windows environments

marcelotduarte:
To improve security.


TTimo:
While "more security" is generally good, this silently breaks a lot things and causes regressions, creating behavior differences between the packaged app and running from the interpreter. Anything using shutil.which is broken now.

Maybe this is something that should default to off, and only be enabled by default for fresh projects. It'll cause a lot of strife.

marcelotduarte:
Open an issue, we will try to resolve both issues as best we can.

marcelotduarte:
Issue #1674

PR #1672, opened 2022-10-29
Include copy of cx_Freeze license with frozen applications

cainesi:
The frozen application contains bootstrap code from cx_Freeze (in order to initialize python and launch the frozen script), so the license terms require that the license be included with the frozen program.  To make that easy for users, we can do that automatically.


cx_Freeze/freeze_license/cx_frozen_license.txt
```diff
@@ -0,0 +1,65 @@
+Why this file is included
```
marcelotduarte: I think you can put this file as doc/src/frozen_licence.txt along with license.rst

cx_Freeze/freezer.py
```diff
@@ -279,6 +279,18 @@ def _freeze_executable(self, exe: Executable) -> None:
             copy_dependent_files=False,
             include_mode=True,
         )
+
+        # copy a file with a the cx_freeze license into frozen application
+        respath = get_resource_file_path(
+            "freeze_license", "cx_frozen_license", ".txt"
+        )
```
marcelotduarte, line 286: ../doc/src

cx_Freeze/freezer.py
```diff
@@ -279,6 +279,18 @@ def _freeze_executable(self, exe: Executable) -> None:
...
+        self._copy_file(
+            respath.absolute(),
+            self.targetdir / "cx_frozen_application_license.txt",
```
marcelotduarte: cx_frozen_application_license -> frozen_application_license or cx_freeze_application_license
cx_freeze/cx_Freeze is the package name

setup.py
```diff
@@ -284,4 +322,7 @@ def _make_strs(paths: List[Union[str, Path]]) -> List[str]:
         cmdclass={"build_ext": BuildBases},
         options={"install": {"optimize": 1}},
         ext_modules=extensions,
+        package_data={'cx_Freeze':
```
marcelotduarte: If you use MANIFEST.in, you can remove the changes in setup.py

marcelotduarte (changes requested):

cx_Freeze/freezer.py
marcelotduarte, line 286: Maybe you can put this txt file in bases or initscripts.

cx_Freeze/freezer.py
cainesi: I'll go with frozen_application_license

cx_Freeze/freezer.py
cainesi: I'll put it in initscripts.

setup.py
cainesi: I've never used MANIFEST.in before, thanks for pointing that out.

marcelotduarte: after
recursive-include cx_Freeze/hooks *.conf *.sh *.qrc
add
recursive-include cx_Freeze/initscripts *.txt

cainesi:
I think everything is resolved now.  Did you want any other changes?

marcelotduarte:
Yes I want. Sorry to bother you, but I think this one is small. The update_frozen_license routine can be placed in a new module, like the one we have in ci/requirements.py and then add the call to .pre-commit-config.yaml

marcelotduarte:
You can enable pre-commit in your environment to reduce the time to wait for pass the commit checks.
See https://cx-freeze.readthedocs.io/en/latest/development/index.html#setup
to check manually: pre-commit run -a -v --hook-stage manual

cainesi:
> Yes I want. Sorry to bother you, but I think this one is small. The update_frozen_license routine can be placed in a new module, like the one we have in ci/requirements.py and then add the call to .pre-commit-config.yaml

No problem.  I've never used the pre-commit mechanism before, so it is a good learning experience.  I have set it up locally now, so that should be a help. 

I'll rework the frozen license update to be more like requirements.py, but I just want to make sure I understand what is going on with that one.   I think the relevant part of .pre-commit-config.yaml is:

```
- repo: local
  hooks:
  - id: requirements
    name: requirements
    language: system
    entry: python ci/requirements.py
    files: ^requirements.*.txt$
```

If I'm understanding right, this is telling to pre-commit to run ci/requirements.py if changes are being made to files matching `^requirements.*.txt$`.  The `requirements.py` script re-makes `requirements.txt` and `requirements-dev.txt` based on what it finds in `config.cfg`.  I am wondering if that hook should instead be triggered by a change to `setup.cfg`?

(And if it should not be changed to setup.cfg, is the expression above correct for matching `requirements-dev.txt`?)

marcelotduarte:
If I do a change in setup.cfg then requirents.txt is updated in the commit.
For the license, we can do the same.

cainesi:
> If I do a change in setup.cfg then requirents.txt is updated in the commit.
> For the license, we can do the same.

Sorry, I think I am missing something.  If setup.cfg changes, what is it that changes requirements.txt?  Is it the "requirements" pre-commit hook?

marcelotduarte:
Test it. Create a branch, change a version of setuptools for example to >60, git commit.

marcelotduarte:
For this case, we can use:
```
- repo: local
  hooks:
  - id: add_frozen_license
    name: add_frozen_license
    language: system
    entry: python ci/frozen_license.py
    files: ^cx_Freeze/initscripts/frozen_application_license.txt$
```

But, I can make this. I can merge it as is, and change it to use pre-commit....

cainesi:
> Test it. Create a branch, change a version of setuptools for example to >60, git commit.

I tried it.  When I just changed setup.cfg, requirements.txt was not changed.  But when I also adjusted the .pre-commit-config.yaml line to be `files: setup.cfg` then requirements.txt was updated.

Take a look at cainesi/cx_Freeze@ea5dff8e (with the existing pre-commit-config) and cainesi/cx_Freeze@a15a85b1 (with the changed pre-commit-config).

Maybe I'm missing something?  I took a look at the pre-commit documentation, but it's pretty unhelpful!

marcelotduarte:
For me, with the actual configuration, when I change the setup.cfg and commit, it update the requirements, or if I run manually with `pre-commit run -a -v --hook-stage manual`

marcelotduarte:
With 2 changes to setup.cfg: 
[pre-commit.txt](https://github.com/marcelotduarte/cx_Freeze/files/9921788/pre-commit.txt)

marcelotduarte (approved):

marcelotduarte:
Thanks! I'll make the changes to use pre-commit.

marcelotduarte:
PR #1676

cainesi:
> For me, with the actual configuration, when I change the setup.cfg and commit, it update the requirements, or if I run manually with `pre-commit run -a -v --hook-stage manual`

I took a look at the pre-commit source code, and I'm almost certain that the `files:` parameter is supposed to be the file that should trigger the hook when it is changed (i.e., `setup.cfg` in our case). 

Relevant code is in pre-commit/commands/run.py.  Basically it starts from the list of git staged filed (generated by _all_filenames()), and then the hook runs if anything in that list (subject to some modifications) matches the `files` parameter (Classifier.filenames_for_hook()).

Not sure why the hook would have run for you with files: set to requirements.txt, if it was only the setup.cfg that was changed.  Maybe you are somehow forcing all hooks to run?

Anyway, thanks for taking care of this.  Hope all is well there.

marcelotduarte:
> I took a look at the pre-commit source code, and I'm almost certain that the `files:` parameter is supposed to be the file that should trigger the hook when it is changed (i.e., `setup.cfg` in our case).

We have to check both files to avoid accepting changes manually in the generated file.
In PR #1690 I fixed it.

cainesi:
> We have to check both files to avoid accepting changes manually in the generated file.
> 
> In PR #1690 I fixed it.

Good point. Thanks.

PR #1793, opened 2023-02-23
Check that parent directory exists before writing to file

sudilav:
I found that my system, which uses importlib, error'd when the parent directory at this line didn't exist, when I added a check and make if not it worked and completed the build.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -610,8 +610,7 @@ class Freezer:
                             parts.append("__init__")
                         target_name = target_lib_dir.joinpath(*parts)
                         target_name = target_name.with_suffix(".pyc")
-                        if not os.path.exists(target_name.parent):
-                            self._create_directory(target_name.parent)
+                        self._create_directory(target_name.parent)
                         target_name.write_bytes(data)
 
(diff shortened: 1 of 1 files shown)


cx_Freeze/freezer.py
```diff
@@ -610,6 +610,8 @@ def _write_modules(self, filename: Path, finder: ModuleFinder):
                             parts.append("__init__")
                         target_name = target_lib_dir.joinpath(*parts)
                         target_name = target_name.with_suffix(".pyc")
+                        if not os.path.exists(target_name.parent):
```
marcelotduarte: Please, change to:
self._create_directory(target_name.parent)

marcelotduarte (changes requested):

cx_Freeze/freezer.py
sudilav: Done

marcelotduarte: You can remove the if. I mean, change the two lines to one.

marcelotduarte (approved):

marcelotduarte:
Thanks!

PR #1860, opened 2023-04-01
Fix code for year 2038

bmwiedemann:
This fix is modeled after https://github.com/python/cpython/commit/bb21e28fd08f894ceff2405544a2f257d42b1354


marcelotduarte:
Thanks!

Issue #933, opened 2021-02-25
PyQt5, PyQtWebEngine, bdist_mac: Cannot find QtWebEngineProcess


mikedarcy:
We have PyQt5 app that uses the PyQtWebEngine (embedded Chromium browser) component, which we freeze with cx_Freeze from a `venv` virtual environment. We successfully frozen this application in the past (on both Windows and Mac) using `cx_Freeze 6.3`, `Python 3.7`, and `PyQt 5.11.3`.

Lately, I have been trying to update our out-of-date application dependencies to use `cx_Freeze 6.5`, `Python 3.8` and `PyQt 5.15.2` but I am getting an app crash when trying to load the part of our application that uses PyQtWebEngine. The error is that the frozen application cannot find a required runtime component, `QtWebEngineProcess`, which is essentially the wrapper of the bundled Chromium engine. See [here](https://doc.qt.io/qt-5/qtwebengine-deploying.html#deploying-qt-webengine-processes) for more information about QtWebEngineProcess.

I believe the problem is somewhere in `setRelativeReferencePaths` during `bdist_mac` processing. The program in the `exe.macosx-10-*` build directory _does_ work correctly, but the one in resulting bundled mac app directory does not.

I've tried this with both Python 3.7 and 3.8 on `MacOSX Catalina 10.15.7` and with various versions of PyQt 5.14 and 5.15. Note that after PyQt 5.11.3, the PyQtWebEngine components are bundled separately and have to be installed from PyPi via `pip install PyQtWebEngine`. I have also tried this with every released version of cx_freeze from 6.4 onward, including `6.6.dev0`.

There is some history of this specific file (QtWebEngineProcess) being a thorn in the side of various Python freezing environments (e.g. `pyinstaller`), so I am not that surprised at this regression popping up for us after trying to jump up so many versions of PyQt at once. I guess I'm hoping someone who knows more about the subtleties of the Mac linking mechanisms like @cainesi can work some magic here.

I created a barebones PyQt5 app and corresponding cx_freeze `setup.py` script that can be used to reproduce this. You can reference it in this gist: https://gist.github.com/mikedarcy/44f385488715dc808dbb9fc8b3f3c60d. Just create a new 3.7/3.8 virtual env and `pip install cx_freeze pyqtwebengine` followed by `python setup.py bdist_mac` then try to run the bundled app. If the window draws and renders some text in the window body, then its working, otherwise it will just crash immediately.

Thanks for your time.

[...]

PR #2025, opened 2023-08-31
Builds pass macOS notarization

### LINKED ISSUE #594: Mac OS bundle are rejected by Apple's notarization process
We create a bundle of our python project using the bdist_mac command from cx_freeze. It seems that Apple now requires bundles to be [notarized](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution) to be able to pass the Gatekeeper on target machine. 

Our app is currently being rejected by the notarisation process seemingly because cx-Freeze places all the files (ie executables, libraries, scripts, data) in the `MacOS` folder. That by itself violates many of rules and recommendations by Apple's [codesigning guide](https://developer.apple.com/library/archive/technotes/tn2206/_index.html). For example, all the data and script files [need](https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG306) to be in `Contents/Resources` and although `codesign` is able to sign the bundle, running the verification (ie `codesign -vvv --strict /path/to/bundle`) fail due to existence of data/script files in code locations.

We've been trying to implement a post-cx-freeze step in our build process that tries to restructure the bundle to conform to requirements by the notarisation process. However, the implementation is very hacky and error-prone to future changes of libraries and python version.

It sounds like a problem that needs to be fixed at cx-Freeze's level of operations.

### LINKED ISSUE #1511: Compile an app on a M1 silicon device
I am trying to create an app with the code:

python setup.py bidst_mac

My app uses TensorFlow, so I followed this tutorial first and it runs well before freezing.
https://www.youtube.com/watch?v=4nY5lDBXdOg

However, when I try to run the .app I receive this error:

`Library not loaded: @rpath/libpython3.9.dylib
  Referenced from: /Users/Desktop/main/build/exe.macosx-11.0-arm64-3.9/app
Reason: tried: '/usr/local/lib/libpython3.9.dylib'`

However, it works when I place the dylibs in the usr/local/lib folder, but as soon as I remove the folder it won't start.
Shouldn't the app also run without the files after compiling? because when I move it to another Mac it won't work, since they haven't the files in their usr/local/lib 

What am I missing here?

johan-ronnkvist:
This PR is intended to resolve [this issue](https://github.com/marcelotduarte/cx_Freeze/issues/594) as its primary purpose, in the course of doing this we are also addressing some other smaller issues that are linked further down as well as extending the functionality of `cx_Freeze`'s signing options.

Our intent with this PR is to focus on ensuring the bundle structure is valid with the minimal impact on the internals of the freezer itself -> we are aware of further optimizations that can be made which are listed below - but we'd prefer not to increase the scope beyond whats necessary.

In order to ensure that application bundles pass macOS notarization the essential changes we had to make included the following:

## Changes:
- Moved license file to `Resources` (The folder `MacOS` may _only_ contain binaries) 
- Clean build directory between builds.
- Perform inside-out codesign of all libs/binary files.
- Added new parameters to control behavior of codesign
    - `codesign-verify` 
    - `codesign-timestamp`
    - `codesign-strict` 
    - `codesign-options`
    - `sptcl-assess`
- We now use `--force` when codesigning in order to replace existing signatures.

Upon making these changes and building + deploying a custom version of cx_freeze internally so our CI/CD could make use of it - Our internal pipeline that builds a .app bundle with cx_Freeze now runs and passes notarization by Apple.

We're submitting this as a draft for now for initial feedback as there are a couple of points worthy of further discussion before finalizing these changes.

### Questions
- The current implementation of the adhoc signing is signing as changes happen. In theory, we think it could be an optimization to replace the current adhoc signing flow with our signing method after creation of the .app bundle. Using an adhoc signature if no codesign identity is provided by the user. How would you like us to proceed?
- Would you like for us to update the existing samples? or create a new sample for this flow?

### Thoughts on future improvements/subsequent PR's
- The license file is today moved as a post-step. Is this something that could be modified inside the freezer that places the license file initially?
- Documentation for contributors could be improved regarding use of environment variables and requirements for CI builds. As well as steps for building locally.
- Even with this PR we are not fully adhering to Apples expectations for .app bundle structure because frameworks are not stored in `/Contents/Frameworks/` (but it's still working 😃)
- Consideration should be given to removing the `codesign-resourcerules` parameter as it's no longer supported on newer version of macOS.
- It might be worth adding validation/asserts when the `MacOS` folder only contains binaries

## Related Issues
- In our testing, this PR also resolved [this issue](https://github.com/marcelotduarte/cx_Freeze/issues/1511)
- While testing we discovered that the Qt SimpleBrowser example does not run on OSX due to missing the `QtWebEngineProcess` executable

The work in this PR has been a co-dev effort of myself and @TechnicalPirate. Feedback would be greatly appreciated @marcelotduarte 🙏. Thank you for your efforts in helping us make this contribution.

### DIFF
--- a/cx_Freeze/command/bdist_mac.py
+++ b/cx_Freeze/command/bdist_mac.py
@@ -8,7 +8,7 @@ import itertools
 import plistlib
 import shutil
 import subprocess
-
+from pathlib import Path
 from setuptools import Command
 
 from cx_Freeze.common import normalize_to_list
@@ -543,14 +543,17 @@ class BdistMac(Command):
         signargs = self._get_sign_args()
 
         print(f"About to sign: '{self.bundle_dir}'")
-        for path in itertools.chain(
-            glob.glob(f"{self.bundle_dir}/Contents/Resources/lib/**/*.so", recursive=True),
-            glob.glob(f"{self.bundle_dir}/Contents/Resources/lib/**/*.dylib", recursive=True),
-            [
-                self.bundle_dir,
-            ]
-        ):
-            self._codesign_file(path, signargs)
+        bundle_dir = Path(self.bundle_dir)
+        files_to_sign = set()
+        for item in bundle_dir.iterdir():
+            if item.is_file():
+                if item.suffix == '':
+                    print(f"Found file without extension: {item}")
+                    files_to_sign.add(item)
+                elif item.suffix in {'.so', '.dylib'}:
+                    files_to_sign.add(item)
+        for item in files_to_sign:
+            self._codesign_file(item, signargs)
 
         print("Finished .app signing")
 
@@ -580,6 +583,8 @@ class BdistMac(Command):
         sign_args.append(file_path)
         subprocess.run(sign_args)  # TODO : Protections around this?
 
+
+        # TODO : prob *don't* want to do this on a file by file basis... but maybe we do ? :D
         if self.codesign_verify:
             verify_args = ["codesign", "-vvv", "--deep", "--strict", self.bundle_dir]
             print("Running codesign verification")
--- a/cx_Freeze/command/bdist_mac.py
+++ b/cx_Freeze/command/bdist_mac.py
@@ -18,8 +18,8 @@ from cx_Freeze.darwintools import (
     changeLoadReference,
 )
 
-from ..exception import OptionError
 from ..darwintools import isMachOFile
+from ..exception import OptionError
 
 __all__ = ["BdistDMG", "BdistMac"]
 
--- a/cx_Freeze/command/bdist_mac.py
+++ b/cx_Freeze/command/bdist_mac.py
@@ -298,6 +298,7 @@ class BdistMac(Command):
 
         for darwin_file in self.darwin_tracker:
             # get the relative path to darwin_file in build directory
+            print(f"Setting relative_reference_path for: {darwin_file}")
             relative_copy_dest = os.path.relpath(
                 darwin_file.getBuildPath(), build_dir
             )
@@ -306,7 +307,7 @@ class BdistMac(Command):
             # bundle.  This is the file that needs to have its dynamic load
             # references updated.
             file_path_in_bin_dir = os.path.join(bin_dir, relative_copy_dest)
-
+            print(f"\tfile_path_in_bin_dir: {file_path_in_bin_dir}")
             # for each file that this darwin_file references, update the
             # reference as necessary; if the file is copied into the binary
             # package, change the reference to be relative to @executable_path
@@ -320,6 +321,7 @@ class BdistMac(Command):
                 # this is the reference in the machO file that needs to be
                 # updated
                 raw_path = reference.raw_path
+                print(f"\t\treference.raw_path: {raw_path}")
                 ref_target_file: DarwinFile = reference.target_file
                 # this is where file copied in build dir
                 abs_build_dest = ref_target_file.getBuildPath()
@@ -329,7 +331,7 @@ class BdistMac(Command):
                     file_path_in_bin_dir,
                     oldReference=raw_path,
                     newReference=exe_path,
-                    VERBOSE=False,
+                    VERBOSE=True,
                 )
 
             applyAdHocSignature(file_path_in_bin_dir)
@@ -400,6 +402,7 @@ class BdistMac(Command):
         )
         self.contents_dir = os.path.join(self.bundle_dir, "Contents")
         self.resources_dir = os.path.join(self.contents_dir, "Resources")
+        self.resources_lib_dir = os.path.join(self.contents_dir, "Resources", "lib")
         self.bin_dir = os.path.join(self.contents_dir, "MacOS")
         self.frameworks_dir = os.path.join(self.contents_dir, "Frameworks")
 
@@ -413,11 +416,23 @@ class BdistMac(Command):
         _, self.bundle_executable = os.path.split(executable)
 
         # Build the app directory structure
-        self.mkpath(self.resources_dir)
-        self.mkpath(self.bin_dir)
-        self.mkpath(self.frameworks_dir)
+        self.mkpath(self.resources_dir)  # /Resources
+        self.mkpath(self.resources_lib_dir)  # /Resources/lib
+        self.mkpath(self.bin_dir)  # /MacOS
+        self.mkpath(self.frameworks_dir)  # /Frameworks
 
+        # Copy to relevent subfolders
+        print(f"Executable name: {executable} - {build_exe.build_exe}")
         self.copy_tree(build_exe.build_exe, self.bin_dir)
+        self.copy_tree(os.path.join(self.bin_dir, "lib"), self.resources_lib_dir)
+        shutil.rmtree(os.path.join(self.bin_dir, "lib"))
+        # Make symlink between contents/MacOS and resources/lib so we can use none-relative reference paths later...
+        origin = os.path.join(self.bin_dir, "lib")
+        relative_reference = os.path.relpath(self.resources_lib_dir, self.bin_dir)
+        print(f"Creating symlink - Target: {origin} <-> Source: {relative_reference}")
+        os.symlink(
+            relative_reference, origin, target_is_directory=True
+        )
 
         # Copy the icon
         if self.iconfile:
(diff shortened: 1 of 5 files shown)


marcelotduarte:
Very good! I'll review the code, your questions and then answer your questions.
I will also do a test later.
I'm going to do a rebase and then I need the errors pointed out by the pre-commit to be corrected. I can do it myself if that's the case.

marcelotduarte:
>     * The current implementation of the adhoc signing is signing as changes happen. In theory, we think it could be an optimization to replace the current adhoc signing flow with our signing method after creation of the .app bundle. Using an adhoc signature if no codesign identity is provided by the user. How would you like us to proceed?

I like optimizations, but I'm a fan of the theory that it has to be working well to optimize later.
 
>     * Would you like for us to update the existing samples? or create a new sample for this flow?

If it's pyqt samples for example, I'd rather just update, putting bdist_mac options, it's even better, because I use them for testing and I can compare the results.
 
>     * The license file is today moved as a post-step. Is this something that could be modified inside the freezer that places the license file initially?

Let's leave it the way you did it for now because even if it changes in the freezer, there isn't a folder equivalent to Resources in the current structure.
 
>     * Documentation for contributors could be improved regarding use of environment variables and requirements for CI builds. As well as steps for building locally.

Sure.

>     * Even with this PR we are not fully adhering to Apples expectations for .app bundle structure because frameworks are not stored in `/Contents/Frameworks/` (but it's still working 😃)

Ok.
 
>     * Consideration should be given to removing the `codesign-resourcerules` parameter as it's no longer supported on newer version of macOS.
 
Where is this parameter?

>     * It might be worth adding validation/asserts when the `MacOS` folder does not contain any binaries

>     * While testing we discovered that the Qt SimpleBrowser example does not run on OSX due to missing the `QtWebEngineProcess` executable

 #933
 
> The work in this PR has been a co-dev effort of myself and @TechnicalPirate. Feedback would be greatly appreciated @marcelotduarte 🙏. Thank you for your efforts in helping us make this contribution.

Thanks!

cx_Freeze/command/bdist_mac.py
```diff
@@ -459,26 +519,130 @@ def run(self):
...
+                    return True
+                chunk = file.read(8192)
+        return False
+
+    @staticmethod
+    def _should_sign(file_path: Path):
+        if file_path.suffix in [".so", ".dylib"]:
+            return True
+        if file_path.suffix == "":
+            return BdistMac._is_binary(file_path)
```
marcelotduarte: https://github.com/marcelotduarte/cx_Freeze/blob/main/cx_Freeze/darwintools.py#L28
Does the function _isMachOFile replace _is_binary?

TechnicalPirate:
Hey thanks for the feedback and initial review. 

It seems most of the questions we had can be done as subsequent PR's / improvements.

The only change we're including now based on your feedback is updating the samples as well as resolving the inline comments :) 

> Where is this parameter?

Re:  `codesign-resourcerules`, it's [here](https://github.com/marcelotduarte/cx_Freeze/blob/main/cx_Freeze/command/bdist_mac.py#L180) - i can't find the official announcement but googling around has plenty of results [talking about its deprecation](https://stackoverflow.com/questions/26459911/resource-rules-has-been-deprecated-in-mac-os-x-10-10) 👍 ( _i imagine this will also be logged as its own issue_ ) 

Ready for final review :)

samples/pyside2/setup.py
```diff
@@ -51,7 +52,22 @@
     "include_files": include_files,
     "zip_include_packages": ["PySide2", "shiboken2"],
 }
-
+options = {
```
marcelotduarte, line 49: 'options' is not being used in setup...

cx_Freeze/command/bdist_mac.py
```diff
@@ -459,26 +520,109 @@ def run(self):
...
+                    check=True,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.STDOUT,
+                )
+                print(
+                    "spctl command's output: "
+                    f"{completed_process.stdout.decode()}"
+                )
+            except subprocess.CalledProcessError as error:
+                print(f"spctl check got an error: {error.stdout.decode()}")
```
marcelotduarte, line 627: The correct here is stdout or stderr?

marcelotduarte (changes requested):

marcelotduarte:
> Re: `codesign-resourcerules`, it's [here](https://github.com/marcelotduarte/cx_Freeze/blob/main/cx_Freeze/command/bdist_mac.py#L180) - i can't find the official announcement but googling around has plenty of results [talking about its deprecation](https://stackoverflow.com/questions/26459911/resource-rules-has-been-deprecated-in-mac-os-x-10-10) 👍 ( _i imagine this will also be logged as its own issue_ )

ok, in another issue/pr

Should be the code about [prepare_qt_app](https://github.com/marcelotduarte/cx_Freeze/blob/main/cx_Freeze/command/bdist_mac.py#L375) changed to support newer qt versions or removed (since we no longer support qt4)?

johan-ronnkvist:
> Should be the code about [prepare_qt_app](https://github.com/marcelotduarte/cx_Freeze/blob/main/cx_Freeze/command/bdist_mac.py#L375) changed to support newer qt versions or removed (since we no longer support qt4)?

From what I can see, this code effectively does nothing today. We didn't make any changes here in order to minimize changes. 

The purpose of the `qt_menu.nib` is to ensure native macOS menu bar integration. But it's not something we've come across in testing. 

In my opinion, the code should be removed.

samples/pyside2/setup.py
johan-ronnkvist: Thank you, will fix!

cx_Freeze/command/bdist_mac.py
johan-ronnkvist: The current usage is correct - in this case we should be using `stdout`

marcelotduarte:
> The purpose of the `qt_menu.nib` is to ensure native macOS menu bar integration. But it's not something we've come across in testing.
> 
> In my opinion, the code should be removed.

I also thought about removing it, but I understand from your answer that it is also possible to adapt to qt5/6, right?

marcelotduarte (approved):

marcelotduarte:
Feel free to make new contributions.
@johan-ronnkvist @TechnicalPirate Thank you!

TechnicalPirate:
> Feel free to make new contributions. @johan-ronnkvist @TechnicalPirate Thank you!

You're very welcome - We will use the next dev release to verify everything is still good our side and start making some other contributions as we phase out our wrapper :)

johan-ronnkvist:
> > The purpose of the `qt_menu.nib` is to ensure native macOS menu bar integration. But it's not something we've come across in testing.
> > In my opinion, the code should be removed.
> 
> I also thought about removing it, but I understand from your answer that it is also possible to adapt to qt5/6, right?

Yes, it's technically possible to update these paths to match newer versions of Qt. But I'd suggest to find a valid repro case before doing so. Again, with at least some effort, I wasn't able to find a trace of the mentioned file with newer versions of Qt. I've spent a couple of years working with an application that was cross-platform and never come across any issues with menu bar management in macOS. I can't guarantee it's not relevant, but have not seen usage or the need for it previously.

marcelotduarte:
> Yes, it's technically possible to update these paths to match newer versions of Qt. But I'd suggest to find a valid repro case before doing so.

As there are no issues and given your experience, I agree that it should be removed, even to simplify the code.

marcelotduarte:
> We will use the next dev release to verify everything is still good our side

You can test the patch in the latest development build:
`pip install --upgrade --pre --extra-index-url https://marcelotduarte.github.io/packages/ cx_Freeze`
For conda-forge the command is:
`conda install -y --no-channel-priority -S -c https://marcelotduarte.github.io/packages/conda cx_Freeze`

marcelotduarte:
> > ```
> > * Documentation for contributors could be improved regarding use of environment variables and requirements for CI builds. As well as steps for building locally.
> > ```
> 
> 
> Sure.

PR #2033
https://cx-freeze--2033.org.readthedocs.build/en/2033/development/index.html#building-binary-wheels

mathiascode:
It seems like `Contents/MacOS/lib` and `Contents/Resources/lib` are duplicated in the final .dmg file instead of symlinked. Is this intended?

marcelotduarte:
@mathiascode 
You can test the patch in the latest development build:
`pip install --upgrade --pre --extra-index-url https://marcelotduarte.github.io/packages/ cx_Freeze`
For conda-forge the command is:
`conda install -y --no-channel-priority -S -c https://marcelotduarte.github.io/packages/conda cx_Freeze`

mathiascode:
@marcelotduarte I don't have a macOS installation right now to run the .dmg, but `Contents/MacOS/lib` is a symlink now. Thanks!

marcelotduarte:
> > * While testing we discovered that the Qt SimpleBrowser example does not run on OSX due to missing the `QtWebEngineProcess` executable
> 
> 
> #933

@johan-ronnkvist Please see https://github.com/marcelotduarte/cx_Freeze/issues/933#issuecomment-1799607182

PR #2106, opened 2023-10-20
bdist_mac: move set_relative_reference_paths to build_exe

marcelotduarte:
The focus of this change is the separation of stages, and the result is that it solves some problems. Placing set_relative_reference_paths in the build_exe step corrects all rpaths without sharing a 'freezer' object between build_exe and bdist_mac and thus build_exe is complete. After that, samples (pyside2, pyqt5, etc.) that previously gave an error if run using just build_exe, are now run. And with bdist_mac they also run. To use codesign and notarization you will need the bdist_mac/bdist_dmg step.

This PR depends on #2107

### DIFF
--- a/cx_Freeze/command/bdist_mac.py
+++ b/cx_Freeze/command/bdist_mac.py
@@ -11,8 +11,6 @@ from setuptools import Command
 
 from cx_Freeze.common import normalize_to_list
 from cx_Freeze.darwintools import (
-    DarwinFile,
-    DarwinFileTracker,
     applyAdHocSignature,
     changeLoadReference,
     isMachOFile,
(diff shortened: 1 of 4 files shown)


marcelotduarte:
@TechnicalPirate Can you or @johan-ronnkvist check if this does not break your code/project?
For me, it worked flawlessly using some samples.

TechnicalPirate:
Hi Marcel,

I'd like to look at this but I'm sadly no longer assigned to the project/team I was so have lost access to the test-infrastructure i was using.  😢 I hope i can recreate some of it in my new role...

If i get a window of opportunity to run these tests i will and report back - but right now i can't promise anything.

marcelotduarte:
Alright, Joe. Success in the new role... Marcelo.

Issue #2376, opened 2024-05-01
cx_Freeze with torch.multiprocessing using wrong source in child processes


dmagee:
**Prerequisite**
 - Make sure no duplicated issue has already been reported. You should look for closed issues, too.
 - Make sure you are not asking us to help to solve your specific issue. GitHub issues are opened mainly for development purposes. If you want to ask someone to help to solve your problem, go to some community site like StackOverflow, etc.
 - Make sure your problem is not derived from packaging (e.g. Homebrew).

**Describe the bug**
On linux when I use cx_Freeze with a python script that uses torch.multiprocessing to create multiple threads (which essentially calls multiprocessing) the child processes seem to try to use the original python files (for the program) and original python environment (for python modules), not the ones in the build directory. The initial result of this is errors about the program source .py files not being found. Other errors can occur if the source is copied into the build folder.

**To Reproduce**
Environment is linux, python 3.11, pytorch v2.2.2+cu121 [Note: This problem does not occur on windows]

Minimal source (Minimal.py):
```
import os
os.environ['KERAS_BACKEND']="torch"


import torch

def per_device_launch_fn(current_gpu_index, num_gpu):

    for i in range(1,1000):
        print("Train...")

num_gpu =4

if __name__ == "__main__":

    print("Starting multiprocessing:"+str(num_gpu))
    torch.multiprocessing.start_processes(
                    per_device_launch_fn,
                    args=(num_gpu,),
                    nprocs=num_gpu,
                    join=True,
                    start_method="spawn",
            )

```

build script is

```
import sys
from cx_Freeze import setup, Executable

import sys
sys.setrecursionlimit(5000)

import os
os.environ['KERAS_BACKEND']="torch"

build_exe_options = {"packages": ["onnx","numpy","torch","PIL", "torchvision","keras","sympy","integr
als","multiprocessing"]}

setup(name="Mimimal",version="1.0",description="Minimal",options={"build_exe": build_exe_options},exe
cutables=[Executable("Minimal.py")])
```

**Expected behavior**
I would expect the pyc versions of code in the build folder to be used under all circumstances (even by child processes), not the original ones.

**Desktop (please complete the following information):**
 - Platform information (e.g. Ubuntu Linux 22.04): Linux (not sure version, it's an HPC)
 - OS architecture (e.g. amd64): intel64
 - cx_Freeze version [e.g. 6.11]:6.15.16
 - Python version [e.g. 3.10]: 3.11

[...]

PR #2443, opened 2024-06-07
hooks: fix #2382 regression / improve tests and docs

### LINKED ISSUE #2433: Version 7.1.0 (and 7.1.0.post0) break FastAPI/hypercorn
**Describe the bug**
When running a program built with cx_freeze starting version 7.1.0, it adds additional arguments, which hypercorn does not accept. These arguments seem to be related to multiprocessing integration.

This causes programs built with 7.1.0 or up, to no longer be functional.

A temporary workaround is to downgrade back to 7.0.0.

**To Reproduce**
All that is required is to run a hypercorn/FastAPI server (or it seems anything that uses multiprocessing) built with cx_Freeze on version 7.0.0 or 7.1.0.post0 (and also 7.1.0 was affected).

I added a full example to a reproduction repository:
https://github.com/Daniel-I-Am/cx_freeze-fastapi-repro

**Expected behavior**
I would expect programs that can be built with 7.0.0 to also be able to be built with 7.1.0.

 - Platform information: Docker `python:3.12-alpine` on linux
 - OS architecture: amd64
 - cx_Freeze version: 7.1.0.post0
 - Python version: 3.12

### LINKED ISSUE #2435: Multiprocessing Manager Class Causes Infinite Loop on Mac OS (intel)
**Bug Description**
When I try to create a `multiprocessing.Manager()` object, even with `freeze_support`, it causes an infinite loop.

**Steps To Reproduce**
Create a `main.py` file.
``` python
is_darwin = True # This is determined by a module that does not affect the results.

def main():
    ...

if __name__ == "__main__":
    if is_darwin:
       import multiprocessing; multiprocessing.freeze_support()
        
       print("Creating Managers")
       manager_1 = multiprocessing.Manager()
       print("Finished Creating Manager #1")
       manager_2 = multiprocessing.Manager()
       print("Finished Creating Manager #2")
       manager_3 = multiprocessing.Manager()
       print("Finished Creating Manager #3")
       print("Finished Creating ALL Managers.")

    main()
```

Create a `setup.py` file.

``` python

build_exe_options = {
    ["zip_include_packages"]: []
}

setup(
    name="main-demo",
    version="0.0.0",
    options={"build_exe": build_exe_options},
    executables=[Executable("main.py", base="console")]
)
```

Run the setup file with the following configuration:
``` bash
python setup.py bdist_mac
```

Run the generated application file:
``` bash
./build/main-demo-1.0.0.app/Contents/MacOS/main
```

**Expected behavior**
From the demo, it prints "Creating Managers" repeatedly indefinity.

**System Information:**
 - Platform information: `Mac OS (Sonoma 14.4)`
 - OS architecture: `x86_64`
 - cx_Freeze version: `7.1.0.post0`
 - Python version: `3.12.3`

**Additional context**
Within the project, I am using managers to share data between three separate processes. These processes run a loop attached to a `manager.Event()` object. In the loop, it checks if a boolean value within a `manager.dict()` object is True. If it is, it will call a `tkinter` function; these consist of `filedialog.askdirectory()`, `messagebox.showerror()`, and `simpledialog.askstring()`. Then, it sets the output to a value within the manager dictionary. The process is then repeated with a few other steps. If you would like, I can post the full source code, but currently, the issue only remains with creating the first manager object. I have gotten this to work with `PyInstaller`, but I need to create a `universal2` application, and `cx_Freeze` has been the closest I have gotten to finally completing everything I need.

Although it states in the official documentation that `multiprocessing.freeze_support()` only works on Windows, it has fixed this same infinite loop within `PyInstaller` when I was trying to get that to work.

**Disclaimer**
If there are any misspellings within the code above, that is an accident and not in my actual code. I wrote this issue from my Windows machine all by hand.

marcelotduarte:
Closes #2435 
Closes #2433 
#2376

### DIFF
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -206,10 +206,10 @@ sources.
 Multiprocessing support
 -----------------------
 
-On Linux and macOS, multiprocessing support is automatically managed by
-cx_Freeze, including supporting it in pyTorch.
+On Linux, macOS and Windows, multiprocessing support is managed by cx_Freeze,
+including support for pyTorch.
 
-However, to produce a Windows executable, you must use
+However, to produce an executable, you must use
 `multiprocessing.freeze_support()`.
 
 One needs to call this function straight after the if __name__ == '__main__'
@@ -228,10 +228,11 @@ line of the main module. For example:
         freeze_support()
         Process(target=f).start()
 
-If the freeze_support() line is omitted then trying to run the frozen
-executable will raise RuntimeError.
+If the freeze_support() line is omitted, then running the frozen executable
+will raise RuntimeError on Windows. On Linux and macOS a similar message is
+shown but cx_Freeze tries to run the program by injecting a freeze_support.
 
-Calling freeze_support() has no effect when invoked on any operating system
-other than Windows. In addition, if the module is being run normally by the
-Python interpreter on Windows (the program has not been frozen), then
-freeze_support() has no effect.
+Contrary to what is stated in the Python documentation, freeze_support() is
+required on all operating systems. In addition, if the module is being run
+normally by the Python interpreter on any OS (the program has not been frozen),
+then freeze_support() has no effect.
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -232,7 +232,9 @@ If the freeze_support() line is omitted, then running the frozen executable
 will raise RuntimeError on Windows. On Linux and macOS a similar message is
 shown but cx_Freeze tries to run the program by injecting a freeze_support.
 
-Contrary to what the Python docs may state, you MUST use `multiprocessing.freeze_support()` on MacOS, Windows, and Linux. On Mac and Linux, cx_Freeze patches the call to also handle `multiprocessing.spawn.freeze_support()` when needed.
-required on all operating systems. In addition, if the module is being run
-normally by the Python interpreter on any OS (the program has not been frozen),
-then freeze_support() has no effect.
+Contrary to what the Python docs may state, you MUST use
+`multiprocessing.freeze_support()` on MacOS, Windows, and Linux. On Mac and
+Linux, cx_Freeze patches the call to also handle
+`multiprocessing.spawn.freeze_support()` when needed. In addition, if the
+module is being run normally by the Python interpreter on any OS
+(the program has not been frozen), then freeze_support() has no effect.
--- a/doc/src/faq.rst
+++ b/doc/src/faq.rst
@@ -206,7 +206,7 @@ sources.
 Multiprocessing support
 -----------------------
 
-On Linux, macOS and Windows, multiprocessing support is managed by cx_Freeze,
+On Linux, macOS, and Windows, multiprocessing support is managed by cx_Freeze,
 including support for pyTorch.
 
 However, to produce an executable, you must use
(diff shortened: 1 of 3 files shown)


doc/src/faq.rst
```diff
@@ -228,10 +228,11 @@ line of the main module. For example:
...
-executable will raise RuntimeError.
+If the freeze_support() line is omitted, then running the frozen executable
+will raise RuntimeError on Windows. On Linux and macOS a similar message is
+shown but cx_Freeze tries to run the program by injecting a freeze_support.
 
-Calling freeze_support() has no effect when invoked on any operating system
-other than Windows. In addition, if the module is being run normally by the
-Python interpreter on Windows (the program has not been frozen), then
-freeze_support() has no effect.
+Contrary to what is stated in the Python documentation, freeze_support() is
```
ntindle: I think this could be clarified a bit. The docs (python) are a bit unclear because the need to cause freezesupoort depends on the method to create new processes which widely varies by platform.  Imo I’d mention the freeze from spawn and from base multiprocessing here even if handled by the hook

marcelotduarte: What was done in the hook was to normalize the use of OS-independent freeze_support (monkey-patching it with spaw.freeze_support) and I think it would confuse some people as it would seem like double information or a choice between them depending on the method, with in any case, spaw.freeze_support will always be used.
But how would you write this paragraph?

ntindle: Fair enough, is there one needed for the third type of processing: forkserver

ntindle: There's not

ntindle: It reuses the spawn one if I'm reading correctly

marcelotduarte: It works for forkserver, due to an error I didn't enable the test for forkserver.

ntindle:
```suggestion
Contrary to what the Python docs may state, you MUST use `multiprocessing.freeze_support()` on MacOS, Windows, and Linux. On Mac and Linux, cx_Freeze patches the call to also handle `multiprocessing.spawn.freeze_support()` when needed.
```

ntindle: This isn't my best work of literature, but I think it is more clear what the patch is doing in case someone stumbles across it later

marcelotduarte: Due to the characteristics of forkserver, only the sample3 test can be used with it.
The test pass: https://github.com/marcelotduarte/cx_Freeze/actions/runs/9411907451/job/25925953665?pr=2443#step:6:286

marcelotduarte:
Please do a test without freeze_support so you can see the message (before it was in a loop).

ntindle:
Won't be able to test today unfortunately, Just realized its almost 1am

ntindle:
Works on windows

ntindle:
testing mac next

ntindle:
On mac, when commenting out `multiprocessing.freeze_support()` 5 of the 7 multiprocessing tests still pass
With the `freeze_support` enabled, all tests pass. 

The failing tests are sample1-spawn and sample2-spawn. They fail in expected ways. The passing tests have no output but that may be a config issue on my side (or false passes)

marcelotduarte:
> The failing tests are sample1-spawn and sample2-spawn. They fail in expected ways.

They fail but do not enter an infinite loop as before, right?

ntindle:
Correct

PR #2444, opened 2024-06-07
bdist_rpm: Fix string concat error due to order of op for + and or in RPM

ntindle:
# Background

When running an rpm build, if you don't have self.distribution.get_description populated, cx_Freeze should use 'UNKNOWN'. However, due to an order of operations bug, it attempts to concat str with None which isn't possible

# Changes
- Reformat the string concat to an f-string so that "or" is evaluated before the concat


marcelotduarte:
Thanks!

PR #2680, opened 2024-11-12
Fix namespace package containing extensions

skilkis:
Summary of Changes:
- Handle branch when root is a namespace package
- Fix annotation in `stub_code`
- Add test case to demonstrate issue


skilkis:
@marcelotduarte thought it was quickest to contribute this small fix. If you run the test case against the `main` branch it demonstrates the error when `self.root.file` is `None`. Can create an issue report if you'd like to see the traceback 😊

marcelotduarte:
Hi! Can you adapt the test case to use embedded code like [here](https://github.com/marcelotduarte/cx_Freeze/blob/main/tests/test_command_build_exe_excludes.py#L19)?

skilkis:
@marcelotduarte I saw your comment too late before pushing, I can do that!

cx_Freeze/module.py
```diff
@@ -234,14 +234,15 @@ def _file_validate(self, filename: Path | str | None) -> Path | None:
...
     def stub_code(self) -> CodeType | None:
-        cache_path: Path = self.cache_path
+        cache_path: Path | None = self.cache_path
         filename = self._file
         if filename is None:
             return None
         ext = "".join(filename.suffixes)
-        if ext not in EXTENSION_SUFFIXES:
+        file = self.root.file
+        if ext not in EXTENSION_SUFFIXES or file is None:
```
skilkis: Even though we don't get the error anymore with this fix, finding the stubs of extension modules inside of a namespace package is still broken. 

We could choose to look inside `self.root.path` when `self.root.file` is `None`, but I am not sure what this attribute contains as it is typed as a list. Essentially we need a way to obtain the correct `source_dir` of the namespace package. Do you think that would be possible?

marcelotduarte: root is a Module, so root.path is the same as Module.path
https://github.com/marcelotduarte/cx_Freeze/blob/main/cx_Freeze/module.py#L187

skilkis: Thanks for catching that bug in the test, pushed a change to be able to still resolve `stub_code` properly

marcelotduarte:
@skilkis This PR fixes #2683?

skilkis:
> @skilkis This PR fixes #2683?

Yes it should! I can test it on a code base when I get home and let you know for sure

skilkis:
@marcelotduarte sorry for the delay, managed to test it out today and it fixes the issue on the code base where I encountered the issue

marcelotduarte:
Thanks!

marcelotduarte:
[Release 7.2.6](https://github.com/marcelotduarte/cx_Freeze/releases/tag/7.2.6) is out!

PR #2871, opened 2025-04-03
bdist_msi: Make MSI checkbox to "launch after install" optional

### LINKED ISSUE #2870: Make "launch on finish" checkbox optional
Version 8.1 added a checkbox to launch the application at the end of the MSI installation. For command line tools that need arguments, this is confusing, since a command prompt would pop up for a few seconds, print a warning that a user doesn't have time to read and then disappear.

Would it be OK to add an option: `launch-on-finish` (default: `False`) to the `bdist_msi` configuration to control if this checkbox should appear or not?

If yes, I can try to do this.

CC @MeGaGiGaGon

jhoekx:
Make the checkbox in the MSI installer to launch the application when the installation finishes optional.

This closes #2870 .

### DIFF
--- a/cx_Freeze/command/bdist_msi.py
+++ b/cx_Freeze/command/bdist_msi.py
@@ -125,8 +125,10 @@ class bdist_msi(Command):
             "rft formatted license file to include in the installer",
         ),
         (
-            "launch-on-finish", None, "Include a Launch on Finish checkbox in the installer."
-        )
+            "launch-on-finish",
+            None,
+            "Include a Launch on Finish checkbox in the installer.",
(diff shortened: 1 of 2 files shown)


MeGaGiGaGon:
Nice catch, I didn't realize this when making my original PR. One thought from me; should the GUI/non arg taking TUI based msi samples have their configs adjusted to have this set to true?

marcelotduarte:
He restored backward compatibility, which is correct. That is, in order to use the new feature, you must indicate that you want to use it.

marcelotduarte:
Thanks!

PR #2998, opened 2025-06-11
startup: Fixed crash when sys.path contains non-str type

1ZUMIKun:
Some code will add non-str types (such as pathlib.Path) to sys.path, which should be ignored at this time according to the official python documentation.But in cx_Freeze this causes a crash, so we need to skip the non-str type.

A program is free to modify this list for its own purposes. Only strings should be added to [sys.path](https://docs.python.org/3.13/library/sys.html#sys.path); **all other data types are ignored during import**.

### DIFF
--- a/cx_Freeze/initscripts/__startup__.py
+++ b/cx_Freeze/initscripts/__startup__.py
@@ -42,6 +42,8 @@ class ExtensionFinder(PathFinder):
             return None
         suffixes = EXTENSION_SUFFIXES
         for entry in sys.path:
+            if type(empty) is not str:
+                continue
             if ".zip" in entry:
                 continue
             for ext in suffixes:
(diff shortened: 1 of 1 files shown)


marcelotduarte:
Hi! Do you have any code examples or usage of a package that makes this change so that I can use it as a test?

1ZUMIKun:
Here's an example.

```
import os
import sys
from pathlib import Path

def get_module_path_list(current_path):
    parent_path = os.path.abspath(os.path.join(current_path, "modules"))
    return [current_path, parent_path]

now_path = Path.cwd()

module_path_list = get_module_path_list(now_path)
sys.path.extend(module_path_list)

import numpy # Or any module
```

marcelotduarte:
Thanks!

marcelotduarte:
[Release 8.4](https://github.com/marcelotduarte/cx_Freeze/releases/tag/8.4.0) is out!

Issue #2531, opened 2024-08-02
PySide6.7.2 + cx_Freeze=7.1.1 + Ubuntu 20.04 very big size of build directory - not used qt plugins and *.so files duplicates many times


break11:
**Describe the bug**

I attach simple Qt App like this:

```
import sys

from PySide6.QtCore import Qt
from PySide6.QtWidgets import QApplication, QLabel, QWidget


def main() -> int:
    app = QApplication(sys.argv)
    window = QWidget()
    window.setWindowTitle("Simple")
    window.setGeometry(300, 300, 300, 300)
    label = QLabel(window)
    label.setText("Hello World!")
    label.setGeometry(0, 0, 300, 300)
    label.setAlignment(Qt.AlignCenter)
    window.show()
    return app.exec()


if __name__ == "__main__":
    sys.exit(main())
```

After build with cx_Freeze the build folder size is 1Gb ! It is to big.
1) Inside build folder i found many folder that not necessary in this app - many qt plugins. After i delete it build folder size became smaller - about 230 Mb. This is quite acceptable. And after deleting - app in build folder still work. I wrote small script for this deletion - it in attachment.

2) Many *.so files in build folder has duplicate - like libicudata.so.73 - it size about 30 Mb - 
first location --- build/exe.linux-x86_64-3.12/lib/PySide6/Qt/lib/ 
second location --- build/exe.linux-x86_64-3.12/lib/PySide6/Qt/plugins/platforms/ 
third location --- build/exe.linux-x86_64-3.12/lib/PySide6/Qt/plugins/xcbglintegrations/

It is 3*30 Mb instead of 30 Mb And there are many such files - like libQt6Core.so.6 (7Mb), libQt6Gui.so.6 (10 MB) and other.

I replace duplicate of some of this files to sym link - and all still work.

If i use simple app with QWebEngine it's even worse. The size of build folder about 1.6 Gb before cleaning.

I try this simple project on Win10 and bug not present there. The size is approximately the same as after cleaning in Ubuntu.

**To Reproduce**
[cx_Freeze.zip](https://github.com/user-attachments/files/16463667/cx_Freeze.zip).

**Expected behavior**
Remove not used Qt plugins, create sym links in build process?

**Desktop:**
 - Ubuntu Linux 20.04:
 - amd64:
 - cx_Freeze  7.1.1:
 - Python version [3.9 - 3.12]:

kristofmulier:
[...]

PR #3036, opened 2025-07-07
finder: detect editable packages (Python 3.10+ only)

frmdstryr:
Currently if a package is installed from a local path in editable mode it is not found by cx-freeze and exits with an ImportError no module named "foo" message. 

When a package is installed in editable mode it creates a module in the site-packages folder `__editable__<dist_name>_<version>_finder.py` that contains an implementation of [MetaPathFinder](https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder) named `_EditableFinder` (at least on 3.10+).

This PR makes cx-freeze's finder try to use that to lookup the ModuleSpec when the normal PathFinder doesn't find a module.

I've tested this locally in a project using several editable packages (that include submodules) and it is working well.. but am not sure where to add a test in cx-freeze (or if this feature is wanted).

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -393,9 +393,7 @@ class ModuleFinder:
                 ):
                     try:
                         mod = importlib.import_module(f.stem)
-                        spec = mod._EditableFinder.find_spec(
-                            name, path
-                        )  # noqa SLF001
+                        spec = mod._EditableFinder.find_spec(name, path)
                         if spec:
                             return spec
                     except (ImportError, AttributeError) as e:
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -369,6 +369,40 @@ class ModuleFinder:
             return None
         return module
 
+    @staticmethod
+    def _find_editable_spec(
+        name: str, path: Sequence[str] | None
+    ) -> importlib.machinery.ModuleSpec | None:
+        """Find the spec for a module installed as an editable package."""
+        if hasattr(importlib.metadata, "packages_distributions"):
+            # the distribution name may vary from the module name (eg may include '-').
+            # packages_distributions returns the mapping but is only available on 3.10+
+            dist_names = importlib.metadata.packages_distributions().get(
+                name, []
+            )
+        else:
+            dist_names = [name]
+
+        for dist_name in dist_names:
+            dist = importlib.metadata.distribution(dist_name)
+            if not dist:
+                continue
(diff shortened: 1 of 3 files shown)


marcelotduarte:
I need tests, even to be able to review.
Add the test to this module:
https://github.com/marcelotduarte/cx_Freeze/blob/main/tests/test_modulefinder.py

cx_Freeze/finder.py
```diff
@@ -369,6 +369,40 @@ def _internal_import_module(
...
+            dist = importlib.metadata.distribution(dist_name)
+            if not dist:
+                continue
+            for f in dist.files:
+                if f.name.startswith("__editable__") and f.name.endswith(
+                    "_finder.py"
+                ):
+                    try:
+                        mod = importlib.import_module(f.stem)
+                        spec = mod._EditableFinder.find_spec(name, path)
```
marcelotduarte: Add a comment `# noqa: SLF001` to pass pre-commit.

marcelotduarte: I made a typo. The correct is `# noqa: SLF001`.
I've already corrected it in your code.

marcelotduarte:
If I understand correctly, these tests should not pass on py39, right? So, limit the test to py310+.

frmdstryr:
Should I skip the test on 3.9? ~~It works if I install the module locally but for some reason it's not working in the test like a something is being cached.~~ Edit: Please disregard that

marcelotduarte:
Thanks!

marcelotduarte:
To test the dev version:
`pip install --extra-index-url https://test.pypi.org/simple/ cx-Freeze==8.4.0.dev1752292930`
Using uv:
`uv pip install --extra-index-url https://test.pypi.org/simple/ cx-Freeze==8.4.0.dev1752292930 --index-strategy unsafe-best-match`

frmdstryr:
Thank you!

marcelotduarte:
[Release 8.4](https://github.com/marcelotduarte/cx_Freeze/releases/tag/8.4.0) is out!

frmdstryr:
Nice, thanks! 

It looks like the duplicate files issue reported in https://github.com/marcelotduarte/cx_Freeze/issues/2531 has returned? I created a workaround here https://github.com/frmdstryr/cx_Freeze/tree/avoid-duplicate-qt-libs but it is only for conda so didn't make a PR...

marcelotduarte:
Please make a PR. Thanks!

PR #3063, opened 2025-08-07
bdist_appimage: add "runtime-file" argument to bdist_appimage command

fklemme:
When using bdist_appimage to build an AppImage, the appimagetool will automatically download the latest "type2 runtime" from https://github.com/AppImage/type2-runtime/releases to be included in the AppImage file. While such an automatic download can be convenient, some environments require full "offline" buildability of the target application. The appimagetool already provides a solution for this with the "--runtime-file" argument, which is not yet used in cx_Freeze.

This patch adds an additional argument "runtime" to the bdist_appimage command, which behaves very similar to the existing argument "appimagekit". If provided, the runtime will downloaded to the given path (if not already there) and it will be passed onwards to the appimagetool when invoked in the freezing process.

I tried to test my approach by following the build instructions in doc/src/development/index.rst and tests/README.md but failed miserably. Please take the suggested changes with care. I will try again in a container. Any suggestions are welcome.

### DIFF
--- a/cx_Freeze/command/bdist_appimage.py
+++ b/cx_Freeze/command/bdist_appimage.py
@@ -30,7 +30,9 @@ __all__ = ["bdist_appimage"]
 
 ARCH = platform.machine()
 APPIMAGETOOL_RELEASES_URL = "https://github.com/AppImage/appimagetool/releases"
+TYPE2RUNTIME_RELEASES_URL = "https://github.com/AppImage/type2-runtime/releases"
 APPIMAGETOOL_DOWNLOAD = f"download/continuous/appimagetool-{ARCH}.AppImage"
+TYPE2RUNTIME_DOWNLOAD = f"download/continuous/runtime-{ARCH}"
 APPIMAGETOOL_CACHE = f"~/.local/bin/appimagetool-{ARCH}.AppImage"
 
 
@@ -44,6 +46,11 @@ class bdist_appimage(Command):
             None,
             "path to appimagetool (formerly AppImageKit) "
             f'[default: "{APPIMAGETOOL_CACHE}"]',
+        ),
+        (
+            "runtime=",
+            None,
+            "path to type2 runtime (optional)",
         ),
         (
             "bdist-base=",
@@ -76,6 +83,7 @@ class bdist_appimage(Command):
 
     def initialize_options(self) -> None:
         self.appimagekit = None
+        self.runtime = None
 
         self.bdist_base = None
         self.build_dir = None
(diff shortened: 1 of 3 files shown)


marcelotduarte:
PRs are welcome!

> When using bdist_appimage to build an AppImage, the appimagetool will automatically download the latest "type2 runtime" from https://github.com/AppImage/type2-runtime/releases to be included in the AppImage file. While such an automatic download can be convenient, some environments require full "offline" buildability of the target application. The appimagetool already provides a solution for this with the "--runtime-file" argument, which is not yet used in cx_Freeze.

As you said, appimagetool will automatically download the latest "type2 runtime".  This is corroborated by the appimagetool [readme](https://github.com/AppImage/appimagetool/commit/75c4f370905790a054714f6302c84486ea8d57ca). 

> This patch adds an additional argument "runtime" to the bdist_appimage command, which behaves very similar to the existing argument "appimagekit". If provided, the runtime will downloaded to the given path (if not already there) and it will be passed onwards to the appimagetool when invoked in the freezing process.

So, I don't see the need to download this file again. If you want to maintain an offline build, you'd need to have both files (appimagetool and runtime) already downloaded and passed as options to bdist_appimage. And then, in this case, the runtime option should be implemented. If the intention is to download to the cache and then use offline, I don't see the benefit of downloading the file through cx_Freeze.

> I tried to test my approach by following the build instructions in doc/src/development/index.rst and tests/README.md but failed miserably. Please take the suggested changes with care. I will try again in a container. Any suggestions are welcome.

In the git version I've been making some changes. You can try, using a venv:
```
git clone https://github.com/marcelotduarte/cx_Freeze
cd cx_Freeze
make install
make tests
```

marcelotduarte:
> So, I don't see the need to download this file again. If you want to maintain an offline build, you'd need to have both files (appimagetool and runtime) already downloaded and passed as options to bdist_appimage. And then, in this case, the runtime option should be implemented. If the intention is to download to the cache and then use offline, I don't see the benefit of downloading the file through cx_Freeze.

This is a bit confusing, but I want to say that we should just add a "runtime" option (and I even prefer "runtime-file" as used in apptoolimage), without downloading anything.

fklemme:
Thank you so much for your feedback - and for your tool. :) I'm happy to give something back, if I can.

Using "runtime-file" over "runtime" makes sense. Probably easier for the user to understand where it comes from and what it does. I will use the new name in the following as well.

Regarding the automatic runtime downloading, I see your point. If the user sets the "runtime-file" argument, they probably have it downloaded already. I rather did it this way to stay consistent with the "appimagekit" argument. Looking at that argument, I see basically three different flows:

1. appimagekit not set (default) - The appimagetool is downloaded to default path and executed from there.
2. `appimagekit ~/not/existing/path/tool.AppImage` - The appimagetool is downloaded and executed from there.
3. `appimagekit ~/already/downloaded/tool.AppImage` - The appimagetool already exists and is executed from there.

For the "runtime-file" argument, the equivalent would be:

1. runtime-file not set (default) - The runtime is implicitly downloaded by the appimagetool.
2. `runtime-file ~/not/existing/path/runtime` - The runtime is downloaded and passed to the appimagetool.
3. `runtime-file ~/already/downloaded/runtime` - The runtime already exists and is passed to the appimagetool.

So I think the bigger question is, does case 2. make sense _in either of these arguments_? I think, if the user provides a path, they probably intend to use an already downloaded appimagetool/runtime. Downloading it automatically could provide some convince but also might also be more prone to typos in the path (i.e. a new path is created instead of giving an error message).

I don't have strong feelings about this, I'm fine to have the runtime-file argument behave differently. I just wonder what behavior might be the most intentional for the user. This is why I mirrored the same behavior. Let me know what you think. :)

Regarding building cx_Freeze locally, I tried the following on a clean & updated Ubuntu 24.04 VM.

```
$ sudo apt install python3-venv git make
$ python3 -m venv .venv
$ . .venv/bin/activate

$ git clone https://github.com/marcelotduarte/cx_Freeze
$ cd cx_Freeze
$ make install

/bin/bash: line 2: pre-commit: command not found
make: *** [Makefile:30: install] Error 127

# from here on, I'm guessing what to do...
$ pip install -r requirements-dev.txt
$ make install
$ make tests

/bin/bash: line 1: pytest: command not found

$ pip install pytest
$ make tests

ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --dist=loadfile -nauto --no-cov
  inifile: /tmp/tmp.pLFiwpZVep/pyproject.toml
  rootdir: /tmp/tmp.pLFiwpZVep
```

I would love to get the cx_Freeze build & tests running on my system. So I'm happy for any tips.

marcelotduarte:
I'll answer you about the runtime later.

Install and run manually the packages:
```
pip install -r requirements-dev.txt
pip install -r tests/requirements.txt
pip install -e.
pre-commit install --install-hooks --overwrite -t pre-commit
pytest --dist=loadfile -nauto -v
```

fklemme:
That worked well, thanks a lot!

fklemme:
@marcelotduarte Is there anything I can help with to move this pull-request forward? Should I change the implementation (no download) for the runtime-file argument?

marcelotduarte:
I see it like this, for the appimagekit argument:
1. appimagekit not set (default) - The appimagetool is downloaded or used from cache
2. appimagekit set to valid path - ok, use it
3. appimagekit set to invalid path - error

For runtime-file, it should be a little different:
1. not set - appimagetool sets it implicitly
2. set to valid path - ok, use it
3. set to invalid path - error

Therefore, I don't see the need for downloading, etc.; it will only add more code for maintenance if there's no good reason.
If you want it to be downloaded to the cache, only to be specified in the command later, it can be downloaded in a single command, i.e., the _get_appimagekit method would download both in one go.

marcelotduarte:
[Release 8.5.0](https://github.com/marcelotduarte/cx_Freeze/releases/tag/8.5.0) is out!

PR #3068, opened 2025-08-11
tests: avoid duplicate qt libs

frmdstryr:
The issue with duplicates lib files from https://github.com/marcelotduarte/cx_Freeze/issues/2531 has returned. 

This adds a fix for linux with libs in site-packages and any libs using conda.

### DIFF
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -1386,7 +1386,8 @@ class LinuxFreezer(Freezer, ELFParser):
         lib_files = self.finder.lib_files
         fix_rpath = set()
         fix_needed = {}
-        site_packages_dir = next(
+        conda_prefix = Path(os.environ["CONDA_PREFIX"]) if IS_CONDA else None
+        site_packages = next(
             (path for path in self.path if path.endswith("site-packages")),
             None,
         )
@@ -1411,13 +1412,22 @@ class LinuxFreezer(Freezer, ELFParser):
                     relative = dependent_source.relative_to(source_dir)
                     # dependency located with source or in a subdirectory
                     dependent_target = target_dir / relative
-                elif site_packages_dir and dependent_source.is_relative_to(
-                    site_packages_dir
+                elif site_packages and dependent_source.is_relative_to(
+                    site_packages
                 ):
                     # put the dependency in the in the target_dir under the same
                     # path as it is in the site-packages folder
-                    relative = dependent_source.relative_to(site_packages_dir)
+                    relative = dependent_source.relative_to(site_packages)
                     dependent_target = self.target_dir / "lib" / relative
+                elif conda_prefix and dependent_source.is_relative_to(
+                    conda_prefix / "lib"
+                ):
+                    # put conda libs where they are in the conda env
+                    rel = dependent_source.relative_to(conda_prefix / "lib")
+                    dependent_target = self.target_dir / "lib" / rel
+                    relative = Path(
+                        os.path.relpath(dependent_target, target_dir)
+                    )
                 else:
                     # put the dependency in target_dir along with the binary
                     # file being copied, unless the dependency has already been
--- a/cx_Freeze/freezer.py
+++ b/cx_Freeze/freezer.py
@@ -1386,15 +1386,6 @@ class LinuxFreezer(Freezer, ELFParser):
         lib_files = self.finder.lib_files
         fix_rpath = set()
         fix_needed = {}
-        conda_prefix = (
-            Path(os.environ["CONDA_PREFIX"])
-            if "CONDA_PREFIX" in os.environ
-            else None
-        )
-        site_packages = next(
-            (path for path in self.path if path.endswith("site-packages")),
-            None,
-        )
         for dependent in self.get_dependent_files(source):
             if not self._should_copy_file(dependent):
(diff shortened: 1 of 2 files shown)


tests/hooks/test_qt.py
```diff
@@ -0,0 +1,92 @@
+"""Tests for hooks for qt"""
+
+from __future__ import annotations
+
+import pytest
+
+from cx_Freeze._compat import IS_MACOS, IS_MINGW, IS_WINDOWS
+
+from ..conftest import IS_CONDA
```
marcelotduarte: _compat has IS_CONDA

tests/hooks/test_qt.py
```diff
@@ -0,0 +1,91 @@
...
+    libs = {}
+    for p in build_lib_dir.glob(f"**/{extension}"):
+        if p.name not in libs:
+            libs[p.name] = []
+        libs[p.name].append(str(p.relative_to(build_lib_dir)))
+    return {k: v for k, v in libs.items() if len(v) > 1}
+
+
+@pytest.mark.venv
+@pytest.mark.skipif(not QT_IMPLS, reason="No Qt libraries installed")
```
marcelotduarte: The test returned:
SKIPPED [1] tests/hooks/test_qt.py:69: No Qt libraries installed
I think it will always be skipped.
As we are using a venv, you do not need to check for imports. You can parameterize with the values directly.

tests/hooks/test_qt.py
```diff
@@ -0,0 +1,91 @@
...
+@pytest.mark.parametrize("qt_impl", QT_IMPLS)
+def test_qt(tmp_package, qt_impl) -> None:
+    """Test if anyio is working correctly."""
+    tmp_package.create(
```
marcelotduarte: Also, you should map the packages to work in MinGW and conda, like [here](https://github.com/marcelotduarte/cx_Freeze/blob/main/tests/hooks/test_crypto.py#L57) and [here](https://github.com/marcelotduarte/cx_Freeze/blob/main/tests/hooks/test_ortools.py#L130).
For instance:
tmp_package.map_package_to_conda["PyQt5"] = "pyqt"

cx_Freeze/freezer.py
```diff
@@ -1375,6 +1375,11 @@ def _post_copy_hook(
         lib_files = self.finder.lib_files
         fix_rpath = set()
         fix_needed = {}
+        conda_prefix = Path(os.environ["CONDA_PREFIX"]) if IS_CONDA else None
```
marcelotduarte: Unfortunately the tests that freeze are failing when using pip due to these changes.

marcelotduarte (dismissed):

cx_Freeze/freezer.py
marcelotduarte: It still fails, I'll try debugging it using pip. But I tested it with conda and noticed that it also fails. Do the tests pass on your machine?

frmdstryr: No it doesn't work anymore locally. 

The original test only ran what was already locally available since it clones the "host" conda env (instead of creating a new one) it looks like the test was dependent on various packages I had installed.

frmdstryr: I figured out one issue causing the built libraries to fail to run.

When installing 'pyside6' with conda there are multiple Qt plugin paths.

Conda's `qt6-main` package creates one in `lib/qt6/plugins` but there is also pyside6 bundled plugins in `lib/python3.13/site-packages/PySide6/Qt/plugins/` so cxfreezes' qt_hooks don't copy everything (in my case the wayland plugins) as it picks the qt6/plugins folder.

marcelotduarte: > The original test only ran what was already locally available since it clones the "host" conda env (instead of creating a new one) it looks like the test was dependent on various packages I had installed.

Yes. You should create a minimal env only with cx_Freeze and its dependencies and test dependencies. Then this base venv is cloned for each test and to not mix dependencies like pyqt with pyside.

marcelotduarte:
I discovered what caused the regression you pointed out. See #3079.
I used your test, and using pip, it passed three tests, except for pyside2. Changing app.exec to app.exec_ also passed three, except for pyqt6.
In conda, I tested using only py39. pyqt5 passed the test using exec. Using exec_ passed both pyqt5 and pyside2. 
Test in conda:
```
========================================================= short test summary info ==========================================================
PASSED tests/hooks/test_qt1.py::test_qt[PyQt5]
PASSED tests/hooks/test_qt1.py::test_qt[PySide2]
FAILED tests/hooks/test_qt1.py::test_qt[PyQt6] - ModuleNotFoundError: -c anaconda pyqt=6
FAILED tests/hooks/test_qt1.py::test_qt[PySide6] - Failed: remains unmatched: 'Hello from Qt!'
================================================== 2 failed, 2 passed in 98.81s (0:01:38) ==================================================
```

In short, do a rebase and leave only the test; it is very useful to prevent this regression from happening again.

marcelotduarte:
Thanks!

marcelotduarte:
[Release 8.4.1](https://github.com/marcelotduarte/cx_Freeze/releases/tag/8.4.1) is out!

frmdstryr:
Thanks!

PR #3095, opened 2025-09-05
Fix Windows permission error

timlassiter11:
# Purpose
Fixes a bug when building executables on Windows that causes a PermissionError. The issue stems from attempting to run `Path.is_dir` on a protected directory. For me this is specifically caused by 'C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps' being in my PATH. 

# Rationale
If we don't have permissions for the directory, it shouldn't be added regardless. Simply suppress the error.

Let me know if you'd prefer if I create an issue for this.

Thanks for the awesome tool!

### DIFF
--- a/cx_Freeze/dep_parser.py
+++ b/cx_Freeze/dep_parser.py
@@ -57,9 +57,10 @@ class Parser(ABC):
         env_path = os.environ["PATH"].split(os.pathsep)
         new_path = []
         for path in self._path + self._bin_path_includes + env_path:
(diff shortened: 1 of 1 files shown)


marcelotduarte:
> Thanks for the awesome tool!

You're welcome! I'm glad you liked it!

marcelotduarte:
[Release 8.4.1](https://github.com/marcelotduarte/cx_Freeze/releases/tag/8.4.1) is out!

PR #3147, opened 2025-10-28
finder: optimized implementation of '_find_editable_spec'

### LINKED ISSUE #3119: Finder enters infinite loop on Windows [Version 8.4.0]
Hi,

I'm not sure where the issue, but with version 8.3.0 (compile time: 5 minutes), everything works fine. However, with versions 8.4.0 and 8.4.1, it looks like the finder goes into an infinite loop. I had to kill the process after waiting for 2 hours. I’ve run several tests to check.

Br,
Egor

egormartiniuc:
Optimized the "find_editable_spec" implementation by refactoring it from a static method to a class method, resulting in better performance.

Moved "importlib.metadata.packages_distributions()" call from the static method to ModuleFinder initialization to avoid rebuilding package distributions for each package file.

### DIFF
--- a/cx_Freeze/finder.py
+++ b/cx_Freeze/finder.py
@@ -88,14 +88,9 @@ class ModuleFinder:
         self._tmp_dir = TemporaryDirectory(prefix="cxfreeze-")
         self.cache_path = Path(self._tmp_dir.name)
         self.lib_files: dict[Path, str] = {}
-        if hasattr(importlib.metadata, "packages_distributions"):
-            # the distribution name may vary from the module name (eg may include '-').
-            # packages_distributions returns the mapping but is only available on 3.10+
-            self.packages_distributions = (
-                importlib.metadata.packages_distributions()
-            )
-        else:
-            self.packages_distributions = None
+        self.packages_distributions = (
+            importlib.metadata.packages_distributions()
+        )
 
     def cleanup(self) -> None:
         self._tmp_dir.cleanup()
@@ -362,11 +357,7 @@ class ModuleFinder:
         self, name: str, path: Sequence[str] | None
     ) -> importlib.machinery.ModuleSpec | None:
         """Find the spec for a module installed as an editable package."""
-        if self.packages_distributions is None:
-            dist_names = [name]
-        else:
-            dist_names = self.packages_distributions.get(name, [])
-        for dist_name in dist_names:
+        for dist_name in self.packages_distributions.get(name, []):
             dist = importlib.metadata.distribution(dist_name)
             if not dist:
                 continue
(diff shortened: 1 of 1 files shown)


cx_Freeze/finder.py
```diff
@@ -88,6 +88,14 @@ def __init__(
         self._tmp_dir = TemporaryDirectory(prefix="cxfreeze-")
         self.cache_path = Path(self._tmp_dir.name)
         self.lib_files: dict[Path, str] = {}
+        if hasattr(importlib.metadata, "packages_distributions"):
+            # the distribution name may vary from the module name (eg may include '-').
+            # packages_distributions returns the mapping but is only available on 3.10+
+            self.packages_distributions = (
```
marcelotduarte: I recently dropped support for Python 3.9 in the next release, so you remove the check and tweak the comment.
The line length should be <80 chars.

cx_Freeze/finder.py
```diff
@@ -350,15 +358,14 @@ def _internal_import_module(
...
     ) -> importlib.machinery.ModuleSpec | None:
         """Find the spec for a module installed as an editable package."""
-        # the distribution name may vary from the module name (eg may
-        # include '-'). packages_distributions returns the mapping
-        dist_names = importlib.metadata.packages_distributions().get(name, [])
-
+        if self.packages_distributions is None:
+            dist_names = [name]
+        else:
+            dist_names = self.packages_distributions.get(name, [])
```
marcelotduarte: I imagine it can be simplified here too.

marcelotduarte (changes requested):

cx_Freeze/finder.py
egormartiniuc: Simplification has been updated.

cx_Freeze/finder.py
egormartiniuc: Greate less code updated

cx_Freeze/finder.py
```diff
@@ -350,16 +353,11 @@ def _internal_import_module(
             return None
         return module
 
-    @staticmethod
     def _find_editable_spec(
-        name: str, path: Sequence[str] | None
+        self, name: str, path: Sequence[str] | None
     ) -> importlib.machinery.ModuleSpec | None:
         """Find the spec for a module installed as an editable package."""
-        # the distribution name may vary from the module name (eg may
```
marcelotduarte, line 358: Keep the comments here.

marcelotduarte (dismissed):

marcelotduarte (approved):

marcelotduarte:
Thank you very much!

marcelotduarte:
[Release 8.5.0](https://github.com/marcelotduarte/cx_Freeze/releases/tag/8.5.0) is out!

PR #3273, opened 2026-03-07
Prevent qthooks.py from placing all libs into PyQt6/libs on linux

frmdstryr:
Currently on linux with cx freeze `8.6.0` every library is getting placed in the PyQt6/lib folder (at least when using PyQt6 or PySide6 installed from conda).

This is because the pattern here catches every `so` file conda env's lib folder.  

The app works either way but I don't think it is expected to do this.

### DIFF
--- a/cx_Freeze/hooks/qthooks.py
+++ b/cx_Freeze/hooks/qthooks.py
@@ -197,7 +197,7 @@ class QtHook(ModuleHook):
         if IS_MINGW or IS_WINDOWS:
             patterns = ["*.dll"]
         else:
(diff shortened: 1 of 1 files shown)


frmdstryr:
To avoid duplicates of all the dependent libs I'm using in a hook for my app. Maybe it could do this by default when using conda?

```py
for source in Path(conda_lib_dir).glob('*.so.*'):
    target = f"lib/{source.name}"
    finder.lib_files.setdefault(source, target)
```

frmdstryr:
I'll just use a patch for my specific case.

